PR #13, opened 2020-04-16
load modules when called from a remote session

xpillons:
when calling the run-tests.sh script from a remote session modules are not loadable
sourcing /etc/profile at the beginning of the test script will allow the modules to be used


jithinjosepkl:
Thanks @xpillons!

PR #18, opened 2020-04-30
disable selinux

mvrequa:
Disable selinux on all CentOS images to avoid thorny security challenges, such as setting up shared /home directories in HPC clusters.


Klaas-:
Please try to use selinux instead of disabling it. This is a security feature in RHEL and all derivates. Disabling it will lead to security findings in compliance testing and security audits.

jithinjosepkl:
Thanks @Klaas- for the note. 

@mvrequa , any thoughts on this? Do you think this will impact CycleCluod?

PR #20, opened 2020-05-05
Adding Checksum Verification

anshuljainansja:
Moved _ubuntu-18.04-hpc_ directory inside _ubuntu-18.x_ to create symmetric hierarchy for ubuntu as that of centos.

Added checksum verification code.

Added code to copy test file.


centos/centos-7.x/centos-7.6-hpc/install.sh
```diff
@@ -22,3 +22,5 @@ set -ex
 # optimizations
 ./hpc-tuning.sh
 
+# copy test file
+./copy_test_file.sh
```
jithinjosepkl, line 26: Do you expect this to be different for different distros? If not, invoking the common copy script should be good enough.

common/verify_checksum.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
jithinjosepkl, line 1: Suggestion: Do you want to combine wget also in this utility? Say, it takes two arguments - download link, and checksum? That way the caller scripts will look cleaner.

common/install_gcc-9.2.sh
```diff
@@ -1,29 +1,33 @@
...
 wget ftp://gcc.gnu.org/pub/gcc/infrastructure/mpc-1.0.3.tar.gz
+../../../common/verify_checksum.sh mpc-1.0.3.tar.gz "617decc6ea09889fb08ede330917a00b16809b8db88c29c31bfbb49cbf88ecc3"
 tar -xvf mpc-1.0.3.tar.gz
 cd mpc-1.0.3
 ./configure && make -j$(nproc) &&  make install
 cd ..
 
 # install gcc 9.2
 wget https://ftp.gnu.org/gnu/gcc/gcc-9.2.0/gcc-9.2.0.tar.gz
+../../../common/verify_checksum.sh gcc-9.2.0.tar.gz "a931a750d6feadacbeecb321d73925cd5ebb6dfa7eff0802984af3aef63759f4"
```
jithinjosepkl, line 30: Suggestion: Does it make sense to have a script in top folder that sets the "root_path" ($topdir) and source essential utilities (like download and check checksum)? This way, each script file doesn't have to do this "../../../" and use ${topdir}/<path> instead.

PR #25, opened 2020-07-08
Adding Lustre Client for CentOS 7.x

anshuljainansja:


centos/centos-7.x/centos-7.6-hpc/install_mellanoxofed.sh
```diff
@@ -7,5 +7,8 @@ tar zxvf MLNX_OFED_LINUX-5.0-2.1.8.0-rhel7.6-x86_64.tgz
 
 KERNEL=( $(rpm -q kernel | sed 's/kernel\-//g') )
 KERNEL=${KERNEL[-1]}
+# Uncomment the lines below if you are running this on a VM
+#RELEASE=( $(cat /etc/centos-release | awk '{print $4}') )
+#yum -y install http://olcentgbl.trafficmanager.net/centos/${RELEASE}/updates/x86_64/kernel-devel-${KERNEL}.rpm
```
jithinjosepkl, line 12: Please add a comment

Klaas-:
Hi,
I caution the usage of epel and the lustre client directly from the internet. If a lot of companies are as restrictive with their internet access as the one I work for this will lead to a lot of firewall rules. We for example only have the microsoft service and ipranges allowed by default. This means http://olcentgbl.trafficmanager.net works, epel mirrors and https://downloads.whamcloud.com won't work. I think I already suggested this to the cyclecloud people longer ago -- mirror all the repositories you need to olcentgbl.trafficmanager.net and that will keep the firewall footprint very limited for HPC :)

Greetings
Klaas

vermagit:
@Klaas- Thanks for the feedback, this is good. We'll look at mirroring (at least the Lustre stuff).
Did you also have handy (or remember) specific examples from CycleCloud? If not, its OK. Point taken.

PR #49, opened 2021-02-16
skip IB eth device config, avoid 30s timeout

mvrequa:
RougueWave/OpenLogic CentOS NetworkManager will dynamically configure all eth network adapters. On the IB device (typically eth1) this will cause a 30s delay as the address update times out and will delay the VM provisioning - affecting TDPR.

### DIFF
--- a/centos/common/network-config.sh
+++ b/centos/common/network-config.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -e
+
+sed -i '/\[main\]/a no-auto-default=*' /etc/NetworkManager/NetworkManager.conf
+
+# update network config on reboot
+mkdir -p /lib/systemd/system/cloud-init-local.service.d/
+cat <<EOF > /lib/systemd/system/cloud-init-local.service.d/50-azure-clear-persistent-obj-pkl.conf
+[Service]
+ExecStartPre=-/bin/sh -xc 'if [ -e /var/lib/cloud/instance/obj.pkl ]; then echo "cleaning persistent cloud-init object"; rm /var/lib/cloud/instance/obj.pkl; fi; exit 0'
+EOF
--- a/centos/centos-7.x/centos-7.6-hpc/install.sh
+++ b/centos/centos-7.x/centos-7.6-hpc/install.sh
@@ -28,6 +28,9 @@ source ./set_properties.sh
 # add udev rule
 $COMMON_DIR/../centos/common/add-udev-rules.sh
 
+# add interface rules
+$COMMON_DIR/../centos/common/network-config.sh
+
 # optimizations
 ./hpc-tuning.sh
 
--- a/centos/centos-7.x/centos-7.7-hpc/install.sh
+++ b/centos/centos-7.x/centos-7.7-hpc/install.sh
@@ -28,6 +28,9 @@ source ./set_properties.sh
 # add udev rule
 $COMMON_DIR/../centos/common/add-udev-rules.sh
 
+# add interface rules
+$COMMON_DIR/../centos/common/network-config.sh
+
 # optimizations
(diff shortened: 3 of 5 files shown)


centos/common/network-config.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+sed -i '/\[main\]/a no-auto-default=*' /etc/NetworkManager/NetworkManager.conf
```
jithinjosepkl, line 4: Thanks @mvrequa  for the PR. Please confirm that this is targeted only for eht0.

centos/common/network-config.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# update network config on reboot
+mkdir -p /lib/systemd/system/cloud-init-local.service.d/
+cat <<EOF > /lib/systemd/system/cloud-init-local.service.d/50-azure-clear-persistent-obj-pkl.conf
+[Service]
+ExecStartPre=-/bin/sh -xc 'if [ -e /var/lib/cloud/instance/obj.pkl ]; then echo "cleaning persistent cloud-init object"; rm /var/lib/cloud/instance/obj.pkl; fi; exit 0'
+EOF
```
jithinjosepkl, line 11: Also, please make sure there are no side effects with AN.

centos/common/network-config.sh
mvrequa: no-auto-default=* changes NetworkManager from opt-out, to opt-in. As it, don't manage the network devices unless the config (from cloud-init) specifically calls for it.
Our partners in AzLinux added code in cloud-init which will configure only interfaces which have address in [IMDS](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/instance-metadata-service?tabs=linux)
https://github.com/canonical/cloud-init/pull/709

VLAN NICs will have an IMDS interface, IB NICs will not. So cloud init will define the 
This will work for all combinations of VLAN nics and IB nics and in any order since IMDS will have the VLAN but not the IB.
Testing included 2 VLAN nics and 1 IB nic configuration.

centos/common/network-config.sh
mvrequa: Tested with Hc with AN enabled, disabled - both showed benefit. No reason to think that AN will impact.

centos/common/network-config.sh
mvrequa: See on 3/2 synthetic VM start goes from ~40s to ~4s (35s faster!).

Date | Service | ct | p50 | p90 | p99
-- | -- | -- | -- | -- | --
2021-02-28 00:00:00.0000000 | AzLinux | 227 | 40352 | 47947 | 51294
2021-03-01 00:00:00.0000000 | AzLinux | 231 | 38769 | 47068 | 49364
2021-03-02   00:00:00.0000000 | AzLinux | 256 | 2553 | 4482 | 5416
2021-03-03 00:00:00.0000000 | AzLinux | 158 | 2537 | 5009 | 14331
2021-03-04 00:00:00.0000000 | AzLinux | 83 | 4423 | 4606 | 13445
2021-03-05 00:00:00.0000000 | AzLinux | 68 | 4434 | 4585 | 4912
2021-03-06 00:00:00.0000000 | AzLinux | 85 | 4408 | 4656 | 10311
2021-03-07 00:00:00.0000000 | AzLinux | 91 | 4435 | 4671 | 7997
2021-03-08 00:00:00.0000000 | AzLinux | 65 | 4424 | 4603 | 4736

jithinjosepkl: Awesome, thanks!

jithinjosepkl (approved):

jithinjosepkl (approved):

PR #59, opened 2021-03-11
Include gpu-topology.sh

vermagit:


jithinjosepkl:
Why do we need this? These are already in: 
please see https://github.com/Azure/azhpc-images/blob/master/common/network-tuning.sh

vermagit:
Oops, missed that :)
I thought this wasn't part of the first build.

PR #61, opened 2021-03-16
Added Ubuntu 20.04

souvik-de:
Added directory structure, dependencies and libraries for Ubuntu 20.04; Moved NVDIA Fabric Manager from common to distro folders; Removed kernel update script from install.sh - vanilla images already have the latest kernel; Updated ReadMe files.


anshuljainansja (approved):
Make sure all the files have the correct permissions set (775 for executables and 644 for readme).

PR #64, opened 2021-03-23
Added nfs utils package

edwardsp:


jithinjosepkl:
Do we need this change for Ubuntu as well?

jithinjosepkl (approved):

vermagit:
nfs-utils is already in those files.

edwardsp:
not needed as nfs-utils is now in the image

PR #86, opened 2021-07-15
hold nvidia-fabricmanager version

samtkaplan:
The cuda drivers are installed via nvidia's ".run" script; whereas, the fabric manager is installed via the deb packages.  This means that if you run `apt-get upgrade`, `apt-get update`, the fabric manager and drivers will not be in sync, and you will get a cuda initialization error.  So, this change tries to fix this issue.


jithinjosepkl (approved):

jithinjosepkl:
Thanks @samtkaplan

PR #90, opened 2021-07-29
IPoIB fix for Ubuntu and CentOS images with waagent v2.3.1.1

abhamidipati0614:


N3WWN:
Does this take into account CentOS systems that have WALinuxAgent already installed as an RPM?

abhamidipati0614:
> Does this take into account CentOS systems that have WALinuxAgent already installed as an RPM?

Yes it does. However, all the IPoIB devices are not listed. That is the reason to use custom waagent v2.3.3.1

N3WWN:
> Yes it does. However, all the IPoIB devices are not listed. That is the reason to use custom waagent v2.3.3.1

Gotcha.  Would it be better to uninstall the RPM, if it is currently installed, and then install waagent from git so that the RPMDB doesn't contain conflicting information about the files?

jithinjosepkl (approved):

PR #95, opened 2021-08-05
Component Versioning for CentOS and Ubuntu images; gdrcopy v2.3 for NVIDIA GPU drivers

abhamidipati0614:


centos/common/hpc-tuning.sh
```diff
@@ -37,3 +37,4 @@ echo "OS.RootDeviceScsiTimeoutPeriod=300" | sudo tee -a /etc/waagent.conf
 echo "OS.MonitorDhcpClientRestartPeriod=60" | sudo tee -a /etc/waagent.conf
 echo "Provisioning.MonitorHostNamePeriod=60" | sudo tee -a /etc/waagent.conf
 sudo systemctl restart waagent
+echo "WAAGENT="$(python /usr/sbin/waagent --version | grep -o "[0-9].[0-9].[0-9].[0-9]" | head -n 1) | sudo tee -a $IMAGE_DIR/component_versions.txt
```
souvik-de, line 40: Please add a new line after this

ubuntu/common/hpc-tuning.sh
```diff
@@ -44,3 +44,4 @@ echo "OS.RootDeviceScsiTimeoutPeriod=300" | sudo tee -a /etc/waagent.conf
 echo "OS.MonitorDhcpClientRestartPeriod=60" | sudo tee -a /etc/waagent.conf
 echo "Provisioning.MonitorHostNamePeriod=60" | sudo tee -a /etc/waagent.conf
 systemctl restart walinuxagent
+echo "WAAGENT="$(python3 /usr/sbin/waagent --version | grep -o "[0-9].[0-9].[0-9].[0-9]" | head -n 1) | sudo tee -a $IMAGE_DIR/component_versions.txt
```
souvik-de, line 47: Please add a new line after this

centos/common/hpc-tuning.sh
abhamidipati0614: added

ubuntu/common/hpc-tuning.sh
abhamidipati0614: added

centos/centos-7.x/centos-7.6-hpc/set_properties.sh
```diff
@@ -3,3 +3,4 @@
 export TOP_DIR=../../..
 export COMMON_DIR=../../../common
 export TEST_DIR=../../../tests
+export IMAGE_DIR=$(pwd)
```
anshuljainansja, line 6: I would prefer dot(.) over pwd, if they behave the same way.

centos/centos-7.x/centos-7.6-hpc/install_mellanoxofed.sh
```diff
@@ -3,6 +3,7 @@ set -ex
 
 MLNX_OFED_DOWNLOAD_URL=https://azhpcstor.blob.core.windows.net/azhpc-images-store/MLNX_OFED_LINUX-5.4-1.0.3.0-rhel7.6-x86_64.tgz
 TARBALL=$(basename ${MLNX_OFED_DOWNLOAD_URL})
+echo "MOFED="$(echo $TARBALL | awk -F- '{print $2 "-" $3}') | sudo tee -a $IMAGE_DIR/component_versions.txt
```
anshuljainansja, line 6: Consider a case where the file name format changes. Someone will update the file name on the line above but might forget to update this. And as far as my understanding goes that line is not gonna throw an error.
Here two lines require changes.

Think about something like this,
```
version =- "5.4-1.0.3.0"
echo $version | sudo tee -a $IMAGE_DIR/component_versions.txt
tarball = "MLNX_OFED_LINUX-$version-rhel7.6-x86_64.tgz"
MLNX_OFED_DOWNLOAD_URL=https://azhpcstor.blob.core.windows.net/azhpc-images-store/$tarball
```
Here, you just have to update one line.

common/copy_component_versions_file.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
anshuljainansja, line 1: What's the benefit of copying to an intermediate file which is $IMAGE_DIR/component_version.txt rather than this final destination file?

centos/centos-7.x/centos-7.6-hpc/install_mellanoxofed.sh
anshuljainansja:
```
sudo tee -a .......
```
This is at multiple places. If you want you can make it a function in common_dir. Here it could be something like,
```
WRITE_VERSION $version
````

PR #101, opened 2021-09-24
yum sync history to remove rpmdb warning

hmeiland:
sync rpmdb to remove warnings from yum


N3WWN:
Nice catch!

I think another consideration should be replacing the use of `rpm` with `yum` when installing packages.  That way the rpmdb doesn't get out of sync in the first place.

jithinjosepkl (approved):

jithinjosepkl:
Thanks @hmeiland

hmeiland:
Yep, although some of that is in the hands of magic install scripts from e.g. Intel...

PR #114, opened 2021-11-24
Add --dkms to NVIDIA driver run file

basnijholt:
Currently, when basing an image on the CentOS-HPC image and one installs any kernel models, the NVIDIA drivers have to be reinstalled.

Installing the drivers with dkms solves this problem.


jithinjosepkl (approved):

jithinjosepkl:
Thanks @basnijholt

PR #116, opened 2021-12-01
updated to use docker CE

JonShelley:
I removed Moby since it does not support all of the docker features and threw an error when I tried to run MLCommons Inferencing tests. I also added a docker configuration change so that it will not fill up the OS image disk and use a different address space than the IB


jithinjosepkl:
Closing after offline discussion.

PR #125, opened 2022-05-06
Add NDv4 health checks

moes1:
I am hoping to add some health checks for NDv4 VMs to the image repository.


tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -0,0 +1,67 @@
...
+
+correctable=$(echo "$row_remaps" | grep "Correctable Error" | cut -d: -f2)
+uncorrectable=$(echo "$row_remaps" | grep "Uncorrectable Error" | cut -d: -f2)
+
+correctable_v=( ${correctable} )
+uncorrectable_v=( ${uncorrectable} )
+
+
+#Identify any GPUs that have experienced more row remaps than allowed.
+for i in $(eval echo {0..$((ngpus-1))}); do
```
jeseszhang1010, line 51: `$(eval echo {0..$((ngpus-1))})`   -> `$(seq 0 $((ngpus-1)))` might be easier.

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -0,0 +1,19 @@
...
+	./dcgmi-check.sh
+        #run 'nvidia-smi -q -d ROW_REMAPPER' and check for problems 
+	./ecc-errors.sh
+        #run ibstat and check all devices are set up as expected
+	./test-ib.sh
+        #test that host to device/device to host bandwidth meet expectations.
+	./test-bandwidth.sh
+        #test nccl using nvlink and ib
+	./test-nccl.sh
+fi
```
jeseszhang1010, line 19: Is it safe to assume the scripts are always invoked under tests/health_checks/NDv4?

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -0,0 +1,74 @@
...
+#Download the host to device and device to host test from superbench v0.5.
+string="https://raw.githubusercontent.com/microsoft/superbenchmark/release/0.5"\
+"/superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu"
+wget -q $string
+
+#Compile the gpu-copy benchmark.
+nvcc -I/usr/include/ -L/usr/lib/x86_64-linux-gnu/ -lnuma gpu_copy.cu -o gpu-copy
+
+#Run the superbench device to host bandwidth test.
+x_htod=$(./gpu-copy --size 134217728 --num_warm_up 5 \
```
jeseszhang1010, line 18: Let's check the return status of the test first. Same as L22.

tests/health_checks/NDv4/test-ib.sh
```diff
@@ -0,0 +1,54 @@
...
+
+ib=200;
+num_r=0
+num_s=0
+num_ps=0
+
+#Check ibstat Rates and make sure they are all 200.
+for i in $(ibstat | grep "Rate:" | cut -d: -f2 | xargs); do 
+	num_r=$((num_r+1));
+	if [ $i -lt $ib ];
```
jeseszhang1010, line 11: Let's put `then` in this line, to be consistent with other places.

tests/health_checks/NDv4/test-ib.sh
```diff
@@ -0,0 +1,54 @@
...
+
+
+if [ $ib -lt 200 ]; then
+	echo "The minimum ib rate observed is $ib, but it should be 200."
+	pass=0
+fi
+
+if [ $num_s -lt $num_r ]; then
+	echo "The ib 'State' is not set to 'Active' for all"\
+		"devices."
```
jeseszhang1010, line 40: No need to have a new line. Same comment for L46.

jeseszhang1010 (commented):
This commit message can be updated a bit, something like

```
tests: Add basic health checks for NDv4 series VMs.

This patch adds a set of tests to make sure some basic aspects of VM are setup correctly and
performing as expected. The tests includes hardware check, dcgmi check, ecc-errors check, 
bandwidth check, ib rate check, nccl bandwidth check. 

One can run all the tests as a whole via run-health-checks.sh, or run each tests individually.
```

tests/health_checks/NDv4/test-nccl.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+#Load the mpi module.
+module load mpi/hpcx
+
+#Lock graphics clocks to 1400.
+sudo nvidia-smi -lgc 1400 > /dev/null
```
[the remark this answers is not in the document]
moes1: Since we are only running the very large size test we don't really need to lock the graphics clock. However, when running smaller size benchmarks it seems like the time for the GPU to spin up can make a significant difference. I've seen that locking the clocks improves performance on the smaller size nccl test runs (at least the ones using ib), and makes their performance more consistent.

tests/health_checks/NDv4/check-hardware.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+#Count the number of gpu-names nvidia-smi outputs.
+ngpus=$(sudo nvidia-smi --query-gpu=name --format=csv,noheader | wc -l)
+
+#Count the number of nics lshw detects.
+nnics=$(sudo lshw -C network | grep -i product | wc -l)
```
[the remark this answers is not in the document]
moes1: Ok, that's a good idea yeah.

tests/health_checks/NDv4/ecc-errors.sh
moes1: Thanks, that is a bit more readable.

tests/health_checks/NDv4/test-ib.sh
moes1: Thanks, that was a mistake!

tests/health_checks/NDv4/test-bandwidth.sh
moes1: Done!

tests/health_checks/NDv4/check-hardware.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
abhamidipati0614, line 2: Just a suggestion to move these files currently under tests/health_checks/NDv4 into tests/health_checks/common and then make a parameterized call with sku name from tests/health_checks/NDv4

This will enable us to extend these health check scripts to other skus.
@jithinjosepkl @moes1 @jeseszhang1010 what do you guys think?

tests/health_checks/NDv4/test-nccl.sh
```diff
@@ -0,0 +1,53 @@
...
+       LD_LIBRARY_PATH=/usr/local/nccl-rdma-sharp-plugins/lib:$LD_LIBRARY_PATH \
+       -mca coll_hcoll_enable 0 -x NCCL_IB_PCI_RELAXED_ORDERING=1 \
+       -x UCX_TLS=tcp -x UCX_NET_DEVICES=eth0 -x CUDA_DEVICE_ORDER=PCI_BUS_ID \
+       -x NCCL_SOCKET_IFNAME=eth0 -x NCCL_DEBUG=WARN \
+       -x NCCL_TOPO_FILE=/opt/microsoft/ndv4-topo.xml \
+       /opt/nccl-tests/build/all_reduce_perf -b4G -f2 -g1 -e 4G)
+
+#select the average bandwidth from the nccl output.
+bw=$(echo "$nccl_nvlink" | grep "Avg bus bandwidth" | \
+	cut -d: -f2 | cut -d ' ' -f 2 | cut -d. -f1)
```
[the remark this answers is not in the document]
moes1: Ah thanks I didn't know about bc.

tests/health_checks/NDv4/test-nccl.sh
```diff
@@ -0,0 +1,53 @@
...
+
+
+#Run a single node nccl test using nvlink.
+nccl_nvlink=$(mpirun -np 8 --bind-to numa --map-by ppr:8:node -x \
+       LD_LIBRARY_PATH=/usr/local/nccl-rdma-sharp-plugins/lib:$LD_LIBRARY_PATH \
+       -mca coll_hcoll_enable 0 -x NCCL_IB_PCI_RELAXED_ORDERING=1 \
+       -x UCX_TLS=tcp -x UCX_NET_DEVICES=eth0 -x CUDA_DEVICE_ORDER=PCI_BUS_ID \
+       -x NCCL_SOCKET_IFNAME=eth0 -x NCCL_DEBUG=WARN \
+       -x NCCL_TOPO_FILE=/opt/microsoft/ndv4-topo.xml \
+       /opt/nccl-tests/build/all_reduce_perf -b4G -f2 -g1 -e 4G)
```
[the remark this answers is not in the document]
moes1: This is a good idea, yeah.

tests/health_checks/NDv4/test-nccl.sh
```diff
@@ -0,0 +1,53 @@
...
+#Run a single node nccl test using infiniBand.
+nccl_ib=$(mpirun -np 8 --map-by ppr:8:node -bind-to numa -H localhost:8 -x \
+       LD_LIBRARY_PATH=/usr/local/nccl-rdma-sharp-plugins/lib:$LD_LIBRARY_PATH \
+       -mca coll_hcoll_enable 0 -x NCCL_IB_PCI_RELAXED_ORDERING=1 \
+       -x UCX_IB_PCI_RELAXED_ORDERING=on -x UCX_TLS=tcp \
+       -x UCX_NET_DEVICES=eth0 -x CUDA_DEVICE_ORDER=PCI_BUS_ID \
+       -x NCCL_SOCKET_IFNAME=eth0 -x NCCL_NET_GDR_LEVEL=5 -x \
+       NCCL_TOPO_FILE=/opt/microsoft/ndv4-topo.xml -x NCCL_SHM_DISABLE=1 \
+       -x NCCL_P2P_DISABLE=1 /opt/nccl-tests/build/all_reduce_perf -b 4G \
+       -f 2 -g 1 -e 4G)
```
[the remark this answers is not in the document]
moes1: Thanks that does make things more readable.

tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+d=0;
```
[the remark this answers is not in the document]
moes1: Thanks!

tests/health_checks/NDv4/dcgmi-check.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+#Turn on persistence mode.
+sudo nvidia-smi -pm 1 > /dev/null
```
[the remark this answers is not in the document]
moes1: I did try to find a way to do this before but it didn't seem like there was an easy way. I will look into it again though as it would be ideal, yeah.

tests/health_checks/NDv4/run-health-checks.sh
moes1: This is a good point. Perhaps I should allow the user to specify a path, but otherwise default to tests/health_checks/NDv4.

tests/health_checks/NDv4/check-hardware.sh
moes1: My thought was that the different skus may need entirely different tests and so it might get confusing to put the tests all in one directory. GPU device to host bandwidth tests will not be applicable to skus without GPUs, for example. But I can see advantages of both ways of doing things.

tests/health_checks/NDv4/test-nccl.sh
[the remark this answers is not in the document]
jeseszhang1010: I see, thanks for your explanation :) 
I would add this as the comment, so others know the idea behind it.

tests/health_checks/NDv4/run-health-checks.sh
jeseszhang1010: User should be able to invoke run-health-checks.sh in any path, as long as they put the correct path. 
Inside run-health-checks.sh, we just need to ensure we can find those test-*.sh (maybe use pushd and popd) and invoke them.

tests/health_checks/NDv4/check-hardware.sh
jeseszhang1010: I'm okay with current folder org now. 
We do need some sort of refactoring to cover more SKUs though. which requires some design work to figure out what the common checks are for different SKUs, and how to organize the different checks for specific SKUs.

jeseszhang1010:
You don't need to add a new commit for address the reviews. You can just update the code and then do force push.

tests/health_checks/NDv4/dcgmi-check.sh
```diff
@@ -0,0 +1,14 @@
...
+
+#Run the basic (r=1) dcgmi diag tests.
+fails=$(dcgmi diag -r 1 | grep "Fail" | wc -l)
```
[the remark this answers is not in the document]
moes1: Good idea, I've been adding timeouts to the external calls that could possibly hang somehow.

tests/health_checks/NDv4/dcgmi-check.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+#get the number of GPUs
+ngpus=$(sudo nvidia-smi --query-gpu=name --format=csv,noheader | wc -l)
+#array to hold status of persistence mode for each gpu
+declare -a pmvals=()
+
+#Find which GPUs have persistence mode off, record that information in pmvals.
+#Enable persistence mode on GPUs that have it disabled.
```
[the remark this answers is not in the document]
moes1: As @abhamidipati0614 suggested, I am trying to keep track of the state each GPU was in before the script executed so we can return the GPUs to the state we found them in. Probably in most situations either all GPUs will have persistence mode enabled, or all will have it disabled, but it's possible users may want to mix settings for some reason. Is there a reason we shouldn't do this?

tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -0,0 +1,66 @@
...
+done
+
+#Query the nvidia-smi output to identify any row remap failures.
+fails=$(echo "$row_remaps" | grep "Remapping Failure Occurred" | cut -d: -f2)
+gpui=0
+
+#Check if any GPU has had a row remap failure.
+for err in $fails; do
+	if [ "$err" != "No" ]; then
+		echo "GPU $gpui indicates it has experienced a remap"\
```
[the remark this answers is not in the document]
moes1: Good idea, the print statement will say that now if the script encounters a pending row remap.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -0,0 +1,88 @@
...
+#Count the GPUs.
+ngpus=$(nvidia-smi --query-gpu=name --format=csv,noheader | wc -l)
+
+#Download the host to device and device to host test from superbench v0.5.
+string="https://raw.githubusercontent.com/microsoft/superbenchmark/release/0.5"\
+"/superbench/benchmarks/micro_benchmarks/gpu_copy_performance/gpu_copy.cu"
+wget -q $string
+
+#Compile the gpu-copy benchmark.
+nvcc -I/usr/include/ -L/usr/lib/x86_64-linux-gnu/ -lnuma gpu_copy.cu -o gpu-copy
```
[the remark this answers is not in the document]
moes1: We could do that, it would make the script a bit simpler. I just thought it would be good to be transparent about exactly what test is being run. Should we switch?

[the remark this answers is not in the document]
jithinjosepkl: yes, i would vote for including the compiled test in the image.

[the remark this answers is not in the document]
moes1: Based on our conversation offline I have made it so that if nvcc is detected the copy_test_file.sh script downloads and compiles the gpu-copy test. Does this work better? 

https://github.com/moes1/azhpc-images/blob/master/common/copy_test_file.sh#L10

moes1:
When the tests all pass this is the current output. Any opinions? 

[sample_output.txt](https://github.com/Azure/azhpc-images/files/8691856/sample_output.txt)

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+check_hardware=$(./check-hardware.sh)
+detect=$(echo "$check_hardware" | grep "Fail:" | wc -l)
```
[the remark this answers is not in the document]
moes1: I have reworked things to use more error checking as you have suggested.

tests/health_checks/NDv4/run-health-checks.sh
moes1: Users can now pass a directory to the run-health-checks.sh script, or it will just assume the health checks are in the default location. It will run wherever you execute from.

tests/health_checks/NDv4/test-nccl.sh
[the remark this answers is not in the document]
moes1: Great! I've added some explanation in the comments near this call.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -0,0 +1,140 @@
...
+		exit 1;
+	fi
+fi
+#Store output from calls passed to catch_error
+output=""
+
+#Count the number of gpu-name nvidia-smi outputs.
+error_smi="**Fail** nvidia-smi failed with error code"
+#Lock graphics clocks to 1400 to eliminate any time for the GPUs to boost.
+#This likely isn't important for performance here, but we will do it anyway
```
jithinjosepkl, line 60: I think we can remove this comment. Clock frequency has a big impact on bandwidth performance.

tests/health_checks/NDv4/test-ib.sh
```diff
@@ -0,0 +1,75 @@
...
+#Store output from calls passed to catch_error
+output=""
+
+ib=200;
+num_r=0
+num_s=0
+num_ps=0
+
+#Check ibstat Rates and make sure they are all 200.
+exec_ibstat="timeout 3m ibstat"
```
jithinjosepkl, line 26: this will fail if AN is enabled :(..

we can use ibstatus here (with the relevant postprocessing).

jithinjosepkl (changes requested):

tests/health_checks/NDv4/test-bandwidth.sh
moes1: Yes, that's fair. Comment removed.

tests/health_checks/NDv4/test-ib.sh
moes1: Thanks, I have switched to ibstatus and modified the postprocessing appropriately.

tests/health_checks/NDv4/check-hardware.sh
```diff
@@ -0,0 +1,51 @@
...
+##This is a test to verify an NDv4 vm is seeing the expected number of GPUs
+##and CPUs.
```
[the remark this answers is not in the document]
moes1: Fixed

jithinjosepkl (approved):

jithinjosepkl (approved):

PR #128, opened 2022-05-31
Moneo

rafsalas19:
Adding scripts to install Moneo. Modified install scripts for ubuntu and centos to run the monitoring/Moneo install scripts.


centos/centos-7.x/centos-7.6-hpc/install.sh
```diff
@@ -37,6 +37,9 @@ $COMMON_DIR/../centos/common/network-config.sh
 # copy test file
 $COMMON_DIR/copy_test_file.sh
 
+# install monitor tools
```
jithinjosepkl, line 40: Have we verified this on CentOS * versions? If not, we can skip CentOS*, as there won't be any new image updates.

PR #137, opened 2022-06-14
Moneo

rafsalas19:
- Relocated the monitor installation script to ubuntu common directory, as it will not be supported in centos
- Added dependencies to install_util files. For ubuntu 18.4 hpc image I upgraded Python to 3.8
- modified the monitor installation script to allow clean install of follow up Moneo dependencies

### DIFF
--- a/ubuntu/common/install_monitoring_tools.sh
+++ b/ubuntu/common/install_monitoring_tools.sh
@@ -6,7 +6,7 @@ set -e
 pip3 install --upgrade pip
 python3 -m pip install ansible
 
-MONITOR_DIR=/opt/azurehpc/monitoring
+MONITOR_DIR=/opt/azurehpc/tools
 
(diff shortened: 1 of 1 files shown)


ubuntu/common/install_monitoring_tools.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+# Dependencies 
+pip3 install --upgrade pip
+python3 -m pip install ansible
+
+MONITOR_DIR=/opt/azurehpc/monitoring
```
jithinjosepkl, line 9: should we name it moneo itself?

rafsalas19: changed to tools directory as discussed.

jithinjosepkl (approved):

PR #145, opened 2022-06-22
Exclude ethernet connections from ib test. 

moes1:
Added check to test-ib.sh to confirm that we are not checking the rate of any connection that is not InfiniBand.


jithinjosepkl:
Can we please clean the commit history for this PR? Looks like a clean rebase will help here.

PR #152, opened 2022-08-25
Several changes made to the NDv4 health checks.

moes1:
1. The run-health-checks.sh script checks for health-checks in the local directory if a directory is not provided.
2. The ecc-errors.sh script now perform a more detailed exploration of ECC errors. An indication is provided when a large number of correctable ECC errors are occuring.
3. The check-hardware.sh script now checks if the correct number of NVME drives are attached.
4. The test-bandwidth.sh script now has the ability to compile bandwidth executable if it is not detected.
5. The test-MaxPayLoad.sh test has been added to indicate if the set MaxPayload matches the max available on the gpus.

### DIFF
--- a/tests/health_checks/NDv4/check-hardware.sh
+++ b/tests/health_checks/NDv4/check-hardware.sh
@@ -18,10 +18,15 @@ catch_error() {
 output=""
 
 #Count the number of gpu-name nvidia-smi outputs.
-gpu_names="sudo timeout 3m nvidia-smi --query-gpu=name --format=csv,noheader"
+gpu_names="timeout 3m nvidia-smi --query-gpu=name --format=csv,noheader"
 error_smi="Fail: nvidia-smi failed with error code"
 catch_error "$gpu_names" "$error_smi"
 ngpus=$(echo "$output" | wc -l)
+
+smi_warnings="timeout 3m nvidia-smi"
+catch_error "$smi_warnings" "$error_smi"
+out_warn=$(echo "$output")
+nwarn=$(echo "$out_warn" | grep "WARNING" | wc -l)
 
 
 #Count the number of nics lshw detects.
@@ -30,8 +35,19 @@ error_nics="Fail: lshw failed with error code"
 catch_error "$find_nics" "$error_nics"
 nnics=$(echo "$output" | grep -i ConnectX-6 | wc -l)
 
+#Count the number of nvmedrives.
+find_nvme="sudo timeout 3m lshw -C storage" 
+error_nvme="Fail: lshw failed with error code"
+catch_error "$find_nvme" "$error_nvme"
+nnvme=$(echo "$output" | grep nvme | grep logical | wc -l)
+
 #Did either test fail?
 passed=1
+if [ $nwarn -ne 0 ]; then
+	passed=0
+	echo "nvidia-smi reported warnings, see output below:"
+	echo "$out_warn" 
+fi
 
 if [ $ngpus -ne 8 ]; then
 	passed=0
@@ -43,9 +59,14 @@ if [ $nnics -ne 8 ]; then
 	echo "$nnics NICs detected with 8 expected."
 fi
 
+if [ $nnvme -ne 8 ]; then
+	passed=0
+	echo "$nnvme NVMEs detected with 8 expected."
+fi
+
 if [ $passed -ne 1 ]; then
-	echo "Fail: Error when detecting GPUs and NICs."
+	echo "Fail: Error when detecting GPUs, NVMEs, and NICs."
 	exit 1;
 else
-	echo "Correct number of GPUs and NICs detected."
+	echo "Correct number of GPUs, NVMEs, and NICs detected."
 fi
--- a/tests/health_checks/NDv4/ecc-errors.sh
+++ b/tests/health_checks/NDv4/ecc-errors.sh
@@ -23,8 +23,12 @@ echo "ECC error check:"
 
 pass=1
 
+DRAM_ECC_THRESHOLD=20000000
+SRAM_ECC_THRESHOLD=10000
+
+
 #query the number of gpus
-exec_names="sudo timeout 3m nvidia-smi --query-gpu=name --format=csv,noheader"
+exec_names="timeout 3m nvidia-smi --query-gpu=name --format=csv,noheader"
 error_smi="**Fail** nvidia-smi failed with error code"
 catch_error "$exec_names" "$error_smi"
 ngpus=$(echo "$output" | wc -l)
@@ -33,6 +37,13 @@ ngpus=$(echo "$output" | wc -l)
 exec_rows="timeout 3m nvidia-smi -q -d ROW_REMAPPER"
 catch_error "$exec_rows" "$error_smi"
 row_remaps=$(echo "$output")
+
+#query nvidia-smi to collect row remap information.
+exec_ecc="timeout 3m nvidia-smi -q -d ECC"
+catch_error "$exec_ecc" "$error_smi"
+ECC_errors=$(echo "$output")
+
+
 #Query the nvidia-smi output to identify any pending row remaps.
 pending=$(echo "$row_remaps" | grep "Pending" | cut -d: -f2)
 gpui=0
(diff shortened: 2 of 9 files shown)


tests/health_checks/NDv4/test-nccl.sh
```diff
@@ -33,7 +33,7 @@ mpi_text="mpirun -np 8 --bind-to numa --map-by ppr:8:node -x LD_LIBRARY_PATH="
 mpi_text+="/usr/local/nccl-rdma-sharp-plugins/lib:\$LD_LIBRARY_PATH"
 mpi_text+=" -mca coll_hcoll_enable 0 -x NCCL_IB_PCI_RELAXED_ORDERING=1"
 
-exec_text="/opt/nccl-tests/build/all_reduce_perf -b1G -f2 -g1 -e 1G"
+exec_text="/opt/nccl-tests/build/all_reduce_perf -c 1 -b1G -f2 -g1 -e 1G"
```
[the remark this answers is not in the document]
moes1: The test itself returns a nonzero exit status if the check fails, so I have been relying on that. But maybe I should also report the output if the check fails.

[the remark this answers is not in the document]
rafsalas19: Ok this makes sense. Its your call. I like the idea of checking check fail but I see why you wrote it this way initially.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,13 +35,251 @@ catch_error() {
...
+    	    {
+    	        cudaEventRecord(start,0);
+    		if(htod){gpuErrorCheck(cudaMemcpyAsync(gpuBuffer[gpu],\
+				hostBuffer[numa],size,cudaMemcpyDefault));}
+    		else{gpuErrorCheck(cudaMemcpyAsync(hostBuffer[numa],\
+				gpuBuffer[gpu],size,cudaMemcpyDefault));}
+    	        cudaEventRecord(stop,0);
+    	        cudaEventSynchronize(stop);
+    
+    	        dummy_kernel<<<1, 64>>>(gpuBuffer[gpu]);
```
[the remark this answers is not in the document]
moes1: The dummy kernel isn't necessary, I just added it out of paranoia that maybe a compiler could get the idea to optimize out the memory copies if it can see that I'm not doing anything with the data. Good catch, I should delete it or add a comment at least...

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,13 +35,251 @@ catch_error() {
...
+    const size_t numBlocks = (numEntries + blocksize - 1) / blocksize;
+
+    const size_t size = numEntries * sizeof(uint8_t);
+
+    for(int gpu=0;gpu<num_gpus;++gpu)
+    {
+        gpuErrorCheck(cudaSetDevice(gpu));
+        gpuErrorCheck(cudaMalloc( (void **) &(gpuBuffer[gpu]), size));
+
+        initialize_ker<<<numBlocks, blocksize>>>(numEntries,gpuBuffer[gpu]);
```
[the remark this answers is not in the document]
moes1: Cuda API calls are serialized on a stream (and this code only uses the default stream) so the memcpy will never be able to execute until the kernel is complete.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,13 +35,251 @@ catch_error() {
...
+__global__ void dummy_kernel(uint8_t * dummy_var)
+{
+}
+
+__global__ void initialize_ker(const size_t numEntries, uint8_t *gpuBuffer)
+{
+    uint64_t index = blockIdx.x * blockDim.x  + threadIdx.x;
+    if ( index < numEntries)
+    {
+        gpuBuffer[index] = index % 256;
```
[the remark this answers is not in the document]
moes1: I should add a comment. I just did this on the GPU because it would be faster.

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -7,12 +7,44 @@
...
+			true
+		else
+			pass=0
+		fi
+	done
+	#if tests are located in script dir use those
+	if [[ $pass -eq 1 ]]; then
+		TEST_DIR=$DIR
+	fi
+	if [[ $pass -eq 0 ]]; then
```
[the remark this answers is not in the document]
moes1: I decided to remove the absolute path so we are no longer assuming anything about the directory structure where the users are running these codes. This way you can just pull the test scripts from github on any machine and run them, and it will work, even if that directory isn't there.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,13 +35,251 @@ catch_error() {
...
+    copy_loop(num_gpus, num_numa, warmup, 0, numEntries, 0, gpuBuffer,\
+		    hostBuffer, ishtod);
+    copy_loop(num_gpus, num_numa, warmup, loops, numEntries, 1, gpuBuffer,\
+		    hostBuffer, ishtod);
+
+    cleanup(num_numa,num_gpus,numEntries,gpuBuffer,hostBuffer);
+}
+EOF
+
+compile_bench="nvcc -lnuma bandwidth.cu -o gpu-copy"
```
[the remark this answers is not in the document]
moes1: During setup we currently build superbench's bandwidth test. If we have built that test this script will run it instead of compiling this bandwidth test, otherwise it will build and run this one. They can take the same inputs and produce similar bandwidth values. 

But good point, I should probably modify the startup so we are always building the same test.

tests/health_checks/NDv4/test-bandwidth.sh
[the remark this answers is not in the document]
rafsalas19: Discussed off line. marking as resolved

tests/health_checks/NDv4/test-bandwidth.sh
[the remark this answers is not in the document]
rafsalas19: great

tests/health_checks/NDv4/test-bandwidth.sh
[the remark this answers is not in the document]
rafsalas19: Yea, its just weird that its being built as part of the test each time. One time is enough.

tests/health_checks/NDv4/run-health-checks.sh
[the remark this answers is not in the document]
rafsalas19: great

tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -73,18 +84,98 @@ uncorrectable_v=( ${uncorrectable} )
...
-	val=$((correctable_v[i] + uncorrectable_v[i]))
-	if [ $val -gt 511 ]; then
+	val=$((uncorrectable_v[i]))
+	if [ $val -gt 512 ]; then
+		pass=0
+		echo "GPU $i has $val uncorrectable row remap errors."
+	fi
+done
+
+uncorrectableVS=$(echo "$ECC_errors" | grep Volatile -A 4 \
```
jeseszhang1010, line 94: uncorrectableVS and uncorrectableVD, curious what does `V` mean here?

tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -23,6 +23,10 @@ echo "ECC error check:"
 
 pass=1
 
+ECC_THRESHOLD=20000000
+SRAM_THRESHOLD=10000
```
jeseszhang1010, line 27: SRAM_ECC_THRESHOLD?

tests/health_checks/NDv4/ecc-errors.sh
```diff
@@ -23,6 +23,10 @@ echo "ECC error check:"
 
 pass=1
 
+ECC_THRESHOLD=20000000
```
jeseszhang1010, line 26: DRAM_ECC_THRESHOLD?

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -7,12 +7,44 @@
...
 if [ "$#" -eq 0 ]; then
-	TEST_DIR="/opt/azurehpc/test/health_checks/NDv4/"
+	pass=1
+	DIR=$(cd `dirname $0` && pwd)
+	for file in "${files[@]}"; do
+		if test -f "$DIR/$file"; then
+			true
+		else
+			pass=0
+		fi
```
jeseszhang1010, line 21: We only need to check non-existent case, then set pass to 0, right?

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -7,12 +7,44 @@
...
+	fi
+	if [[ $pass -eq 0 ]]; then
+		DIR=$PWD
+		pass=1
+		for file in "${files[@]}"; do
+			if test -f "$DIR/$file"; then
+				true
+			else
+				pass=0
+			fi
```
jeseszhang1010, line 35: Same comment as above

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -82,11 +114,18 @@ else
...
+	fi
+        echo "$line"
+
 	if [ $pass -ne 1 ]; then
 		echo "******************At least one health check failed."\
 			"********************"
 	else
-		echo "*********All health checks have passed, everything "\
+		echo "*********All health checks have passed, everything"\
 			"seems ok.***********"
```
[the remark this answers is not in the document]
jeseszhang1010: Just say `All health checks have passed` would be sufficient.

tests/health_checks/NDv4/run-health-checks.sh
```diff
@@ -7,12 +7,44 @@
 #Verify if an argument was passed to the script. Test if the argument is
 #a directory that contains all the expected health checks. If no arguments
 #are passed use the default directory.
```
jeseszhang1010, line 9: Please add a check of number of arguments, then echo script usage if it's incorrect.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,20 +35,259 @@ catch_error() {
...
 #location indicated by the argument. 
 if [ "$#" -eq 0 ]; then
-	CHECK_DIR="/opt/azurehpc/test/health_checks/NDv4/"
+	pass=1
+	DIR=$(cd `dirname $0` && pwd)
+	if test -f "$DIR/gpu-copy"; then
+		true
+	else
+		pass=0
+	fi
```
jeseszhang1010, line 261: only check non-existent case? Same comments for below.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,20 +35,259 @@ catch_error() {
 	fi
 }
 
+create_benchmark() {
```
jeseszhang1010, line 38: If we can download the benchmarks from superbench repo, why would we re-write it here?
```
#string="https://raw.githubusercontent.com/microsoft/superbenchmark/release/"\
#	"0.5/superbench/benchmarks/micro_benchmarks/gpu_copy_performance/"\
#	"gpu_copy.cu"
```

moes1: We have written it here so the health checks can be used even without access to the external internet. Sometimes that may be blocked after machine setup. But that's a good point I should get rid of that comment.

tests/health_checks/NDv4/ecc-errors.sh
moes1: the v refers to the volatile keyword associated with these errors.

tests/health_checks/NDv4/run-health-checks.sh
moes1: yes, that's right. I will change this.

tests/health_checks/NDv4/run-health-checks.sh
moes1: will change this.

tests/health_checks/NDv4/test-bandwidth.sh
moes1: will change this.

tests/health_checks/NDv4/run-health-checks.sh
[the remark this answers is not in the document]
moes1: Fixed this.

tests/health_checks/NDv4/test-MaxPayLoad.sh
```diff
@@ -0,0 +1,20 @@
...
+		| cut -d' ' -f2)
+	DevCtl=$(sudo lspci -vv -s $i | grep DevCtl: -A 2 \
+		| grep MaxPayload | cut -d' ' -f2)
+	if [[ $DevCap -ne $DevCtl ]]; then
+		echo "DevCap and DevCtl MaxPayload do not match for device $i"
+		pass=0
+	fi
+done
+if [[ $pass -ne 1 ]]; then
+	exit 1;
```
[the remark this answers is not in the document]
moes1: Good point.

tests/health_checks/NDv4/check-hardware.sh
```diff
@@ -23,15 +23,31 @@ error_smi="Fail: nvidia-smi failed with error code"
 catch_error "$gpu_names" "$error_smi"
 ngpus=$(echo "$output" | wc -l)
 
+smi_warnings="sudo timeout 3m nvidia-smi"
```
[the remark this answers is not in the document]
moes1: No, and I found a few other places with unnecessary sudos that I have removed.

tests/health_checks/NDv4/ecc-errors.sh
moes1: yeah, that name is clearer.

tests/health_checks/NDv4/ecc-errors.sh
moes1: that name is clearer, i will change it.

tests/health_checks/NDv4/test-bandwidth.sh
```diff
@@ -35,13 +35,251 @@ catch_error() {
...
+        {
+    	    cudaEvent_t start,stop;
+    	    cudaEventCreate(&start);
+    	    cudaEventCreate(&stop);
+            double totaltime = 0.0;
+    	    float localtime = 0.0;
+    	    for(int iloop=0;iloop<warmup+loops;++iloop)
+    	    {
+    	        cudaEventRecord(start,0);
+    		if(htod){gpuErrorCheck(cudaMemcpyAsync(gpuBuffer[gpu],\
```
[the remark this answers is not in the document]
moes1: It just tells the GPU to record a timestamp in that stream so I believe it should not capture the host time, but just to make the code clearer I will move the error checking out of the timed region.

tests/health_checks/NDv4/test-bandwidth.sh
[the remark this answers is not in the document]
moes1: It will only be built if the executable is not already present. The code tests if it exists before calling the function that writes and compiles the executable.

tests/health_checks/NDv4/run-health-checks.sh
moes1: Added this

tests/health_checks/NDv4/test-bandwidth.sh
jeseszhang1010: How about we save gpu_copy.cu directly? Creating a .cu file via Bash (EOF) seems weird to me.

moes1: Yeah, we can certainly do that. I was trying to make each test self-contained, but if that would be clearer, I can move the gpu copy test into its own file.

moes1: Alright I have separated the gpu copy cuda code out to its own file ("gpu-copy.cu"). 

Now i test if the gpu-copy executable exists already. If it does I run it. If the executable does not already exist I test if gpu-copy.cu exists. If the cu file does not exist the test fails with an error message saying it cannot find gpu-copy.cu. If it does exist it is compiled.

jithinjosepkl (approved):

PR #153, opened 2022-08-29
Updated nvidia gpu driver to 510.85.02

moes1:
There are several changes in this update.

    1. The nvidia gpu driver and fabric manager were updated to 510.85.02 to fix an infoROM corruption issue.
    2. The nccl version as updated to 2.14.3-1 to deal with a change in the
       nccl-tests.
    3. The apt installs from the ubuntu install_utils.sh files were pulled out to a ubuntu/common install_utils.sh file.
    Also the RELAXED_ORDERING setting was removed from the run-tests.sh file as it is not aplicable to all SKUs.

### DIFF
--- a/ubuntu/common/install_nvidiagpudriver.sh
+++ b/ubuntu/common/install_nvidiagpudriver.sh
@@ -7,7 +7,7 @@ CHECKSUM=$2
 
 # Reference - https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#ubuntu-installation
 # Install Cuda
-NVIDIA_VERSION="510.73.08"
+NVIDIA_VERSION="510.85.02"
 if [ ${RELEASE_VERSION} == "1804" ]; then CUDA_VERSION="11.6"; else CUDA_VERSION="11-6"; fi
 wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${RELEASE_VERSION}/x86_64/cuda-ubuntu${RELEASE_VERSION}-keyring.gpg 
 mv cuda-ubuntu${RELEASE_VERSION}-keyring.gpg /usr/share/keyrings/cuda-archive-keyring.gpg
@@ -35,7 +35,7 @@ popd
 
 # Nvidia driver
 NVIDIA_DRIVER_URL=https://us.download.nvidia.com/tesla/${NVIDIA_VERSION}/NVIDIA-Linux-x86_64-${NVIDIA_VERSION}.run
-$COMMON_DIR/download_and_verify.sh $NVIDIA_DRIVER_URL "c854bb2dc3368c0127dc08a85bd902f8558a5149085af13d061c612cf06c2913"
+$COMMON_DIR/download_and_verify.sh $NVIDIA_DRIVER_URL "372427e633f32cff6dd76020e8ed471ef825d38878bd9655308b6efea1051090"
 bash NVIDIA-Linux-x86_64-${NVIDIA_VERSION}.run --silent --dkms
 $COMMON_DIR/write_component_version.sh "NVIDIA" ${NVIDIA_VERSION}
 
--- a/ubuntu/common/install_utils.sh
+++ b/ubuntu/common/install_utils.sh
@@ -31,9 +31,11 @@ apt-get -y install numactl \
                    cmake \
                    libnl-3-dev \
                    libnl-route-3-dev \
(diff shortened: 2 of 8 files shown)


jeseszhang1010 (commented):
Please add the reason we update nvidia drivers in commit message. 

It's because it has the fix to infoROM corruption issue, right?

ubuntu/common/install_nvidiagpudriver.sh
```diff
@@ -7,7 +7,7 @@ CHECKSUM=$2
 
 # Reference - https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#ubuntu-installation
 # Install Cuda
-NVIDIA_VERSION="510.73.08"
+NVIDIA_VERSION="510.85.02"
```
jithinjosepkl, line 10: Should we update fabric manager also? [install_nvidia_fabric_manager.sh](https://github.com/Azure/azhpc-images/blob/master/ubuntu/common/install_nvidia_fabric_manager.sh)

Also, are you planning to have a separate PR for the LTS image?

ubuntu/common/install_utils.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 set -ex
 
+
```
abhamidipati0614, line 4: Please remove extra line

ubuntu/ubuntu-18.x/ubuntu-18.04-hpc/install_utils.sh
```diff
@@ -26,49 +26,5 @@ apt-get install -y libgirepository1.0-dev
...
-
-# copy the azcopy to the bin path
-pushd azcopy_linux_se_amd64_10.12.2
-cp azcopy /usr/bin/
-popd
-
-# Allow execute permissions
-chmod +x /usr/bin/azcopy
+#install apt pckages
+$UBUNTU_COMMON_DIR/instal_utils.sh
```
abhamidipati0614, line 30: apt-get update run twice in this file
Line 13 and again in common install_utils.sh. Please correct.

ubuntu/common/install_nvidiagpudriver.sh
moes1: Done, yes I will create a separate PR for the LTS image.

ubuntu/ubuntu-18.x/ubuntu-18.04-hpc/install_utils.sh
moes1: The line 13 apt-get update needs to be there, so I could resolve this by moving the apt-get update out of the common file. But this will mean I need to have an install_utils.sh file for 18.04-LTS that just calls apt-get update and then calls the common install_utils.sh file. Is this alright?

abhamidipati0614: The apt-get update is used for microsoft.prod repo to be available in the list of repos. 
Is python 3.8 also dependent on that? if so, we could move the python installation steps after install_utils

moes1: It looks like we can flip the order so I will do that.

ubuntu/common/install_nvidiagpudriver.sh
moes1: It turns out I needed to update nccl to get the latest version of the nccl-tests to work. Is that acceptable?

ubuntu/common/install_utils.sh
moes1: done

jithinjosepkl (approved):

ubuntu/ubuntu-18.x/ubuntu-18.04-hpc/install_utils.sh
```diff
@@ -10,6 +10,9 @@ cp ./microsoft-prod.list /etc/apt/sources.list.d/
 curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor > microsoft.gpg
 cp ./microsoft.gpg /etc/apt/trusted.gpg.d/
 
+#install apt pckages
+$UBUNTU_COMMON_DIR/install_utils.sh
+
 apt-get update
```
abhamidipati0614, line 16: Do we still need this?

PR #155, opened 2022-09-08
Updating Moneo tag and ansible install

rafsalas19:
Updating Moneo tag and ansible install

### DIFF
--- a/ubuntu/ubuntu-20.x/ubuntu-20.04-hpc/install_utils.sh
+++ b/ubuntu/ubuntu-20.x/ubuntu-20.04-hpc/install_utils.sh
@@ -42,7 +42,6 @@ apt-get -y install numactl \
                    libnl-route-3-dev \
                    net-tools \
                    libsecret-1-0 \
-		   ansible \
 		   python3-pip \
                    dkms \
                    jq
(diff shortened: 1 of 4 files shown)


ubuntu/ubuntu-20.x/ubuntu-20.04-hpc/install_utils.sh
```diff
@@ -42,7 +42,6 @@ apt-get -y install numactl \
                    libnl-route-3-dev \
                    net-tools \
                    libsecret-1-0 \
-		   ansible \
```
jithinjosepkl, line 45: should we have this change to the LTS image also?

rafsalas19: Let me check. I am not sure if we released Moneo for LTS

rafsalas19: So the LTS version doesn't have Moneo. Pretty sure we decided not to include it.

jithinjosepkl (approved):

ubuntu/ubuntu-20.x/ubuntu-20.04-hpc/install_utils.sh
moes1: Do we need ansible in the LTS version? We recently pulled all these apt installs into the ubuntu/common/install_utils.sh file, but if we need to specialize by image version again we can do that...

moes1 (approved):

PR #161, opened 2022-09-20
Fix several issues with health-checks

moes1:
Adding back compilation of gpu-copy executable at vm setup just so the health checks are completely setup when intallation completes. Also fixing a small bug in the ecc-errors script and making an error message created by the test-bandwidth more clear.

### DIFF
--- a/common/copy_test_file.sh
+++ b/common/copy_test_file.sh
@@ -7,4 +7,23 @@ mkdir -p $DEST_TEST_DIR
 
 cp $TEST_DIR/run-tests.sh $DEST_TEST_DIR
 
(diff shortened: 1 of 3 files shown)


jithinjosepkl (approved):

jeseszhang1010 (commented):
NIT: Let's separate the commit title and commit msg by an empty line as follows

```
Fix several issue with health-checks

Adding back compilation of gpu-copy executable at vm setup just so th…
…e health checks are completely setup when intallation completes. Also fixing a small bug in the ecc-errors script and making an error message created by the test-bandwidth more clear.
```

Also, intallation->installation

jeseszhang1010 (approved):

PR #163, opened 2022-09-20
Addressing "Neighbor table overflow" issue.

moes1:
Changing gc_thresh settings to address "Neighbor table overflow" issue. For an example discussing this see this blog post: https://openai.com/blog/scaling-kubernetes-to-2500-nodes/

Values were chosen to match defaults recommended by AKS: https://docs.microsoft.com/en-us/azure/aks/custom-node-configuration#socket-and-network-tuning


moes1:
Removed the sunrpc tcp_max_slot_table_entries fix because it doesn't survive reboots unless we add a service to modprobe sunrpc. Should we add that?

moes1:
I have decided to modify the PR to add in a systemd service to start sunrpc at startup just to show what that would look like.

ubuntu/common/hpc-tuning.sh
```diff
@@ -30,7 +30,37 @@ cat << EOF >> /etc/security/limits.conf
...
+[Install]
+WantedBy=multi-user.target
+EOF
+
+systemctl enable sunrpc_tcp_settings
+systemctl start sunrpc_tcp_settings
+
 
 # Install WALinuxAgent
 apt-get install python3-setuptools
```
rafsalas19, line 66: Why is this done here rather than utils file?

rafsalas19 (approved):

ubuntu/common/hpc-tuning.sh
moes1: This refers to the install of python3-setuptools? That's a good point we probably should move that.

ubuntu/common/hpc-tuning.sh
```diff
@@ -30,7 +30,37 @@ cat << EOF >> /etc/security/limits.conf
...
+echo "net.ipv4.neigh.default.gc_thresh2 = 8192" >> /etc/sysctl.conf
+echo "net.ipv4.neigh.default.gc_thresh3 = 16384" >> /etc/sysctl.conf
+echo "sunrpc.tcp_max_slot_table_entries = 128" >> /etc/sysctl.conf
+
+
+## Systemd service for starting sunrpc and adding setting parameters
+cat <<EOF >/usr/sbin/sunrpc_tcp_settings.sh
+#!/bin/bash
+
+modprobe sunrpc
```
jeseszhang1010, line 43: Does module sunrpc have any dependencies, which need to be loaded first?

moes1: I don't believe so. "modprobe --show-depends sunrpc" just shows: 

sunrpc.ko

ubuntu/common/hpc-tuning.sh
jeseszhang1010: Prefer separate PR, as it's not related to the major change in this PR.

jeseszhang1010 (approved):

PR #226, opened 2023-04-17
Customizations for NDv5

abhamidipati0614:


common/setup_sku_customizations.sh
```diff
@@ -28,6 +28,9 @@ case \$vmSize in
     standard_hb176*v4)
         /opt/azurehpc/customizations/hbv4.sh;;
 
+    Standard_ND96is*_H100_v5)
```
jithinjosepkl, line 31: lower case?

abhamidipati0614: updated

jithinjosepkl (approved):

PR #232, opened 2023-05-02
Update instructions to refer to new URL.

joe-atzinger:
Updated instructions to point to new URL.


yakovdyadkin (approved):
Looks good to me! 

Just to make sure, @abhamidipati0614 are we the only Alma 8.6 image on Az marketplace or do the AlmaLinux guys still have 8.6 hpc images?

Issue #209, opened 2023-03-16
Simpler maintenance for versions and software BOM


petersatsuse:
The versions which got installed by the scripts are distributed in many files. This means a lot of work do provide newer versions.

A central file for the version would help to have less work to change scripts. It would in addition help for new  minor versions for a distribution

A second benefit would be that such a "config" file could act as a software bill of materials, which is needed to have a secure software supply chain.

The idea is to have all variables, versions, urls and checksums in one file, which gets sourced in install.sh.
I have a working prototype for SLE HPC nearly ready.

Here a small snippet how it could look like

```
 --snip--
 # SLE Version
 # it can be generated, but better is to go through this file
 # and manual an set the right versions of the various sections
 # #source /etc/os-release
 # #export SLE_DOTV=${VERSION_ID}
 # #export SLE_MAJOR=${VERSION_ID%.*}
 export SLE_DOTV=15.4
 export SLE_MAJOR=15
 ...
 # azcopy
 export AZVERSION="10.16.2"
 RELEASE_TAG="release20221108"
 #
 export AZTARBALL="azcopy_linux_amd64_${AZVERSION}.tar.gz"
 export AZCOPY_DOWNLOAD_URL="https://azcopyvnext.azureedge.net/${RELEASE_TAG}/${AZTARBALL}"
 
 # HPC-X
 # pls accept the EULA
 export HPCX_VERSION="2.12"
 export HPCX_DOWNLOAD_URL=https://content.mellanox.com/hpc/hpc-x/v${HPCX_VERSION}/hpcx-v${HPCX_VERSION}-gcc-inbox-suse${SLE_DOTV}-cuda11-gdrcopy2-nccl${HPCX_VERSION}-x86_64.tbz
 export HPCX_CHKSUM="bc315d3b485d13c97cd174ef5c9cba5c2fa1fbc3e5175f96f1a406a6c0699bdb"
 ...
 --snip--
```

abhamidipati0614:
@petersatsuse thank you! The development is already in progress and will be tracked as a part of this PR - https://github.com/Azure/azhpc-images/pull/248

LiquidPT:
Spack development was halted, but the work to move component versioning to a single file was added in #328

PR #240, opened 2023-05-11
New SUSE tree with a central config

petersatsuse:
As describe in Issue #209, a central config makes the maintenance much more easy, as module version and also distribution versions can be handled from a central file.

The concept could be also adopted by the other distributions as, many of the seperate scripts already have variables which could be moved to a central file.


abhamidipati0614:
@petersatsuse looks good to me. Thank you for setting up a config file with versions and other metadata. We are already working on implementing something similar to the rest of the distros and will soon be available.

abhamidipati0614 (approved):

PR #242, opened 2023-05-17
support for kvp_client on azhpc-images

abhamidipati0614:
- kvp_client available at /opt/azurehpc/tools
- Minor bug fix for package update on AlmaLinux*


alma/alma-8.x/alma-8.6-hpc/install.sh
```diff
@@ -61,6 +61,9 @@ $COMMON_DIR/install_health_checks.sh
 # copy test file
 $COMMON_DIR/copy_test_file.sh
 
+# copy kvp client file
```
jithinjosepkl, line 64: maybe move this into utils?

common/copy_kvp_client.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+DEST_DIR=/opt/azurehpc
```
jithinjosepkl, line 4: can we use /opt/azurehpc/tools  folder?

alma/alma-8.x/alma-8.6-hpc/install.sh
abhamidipati0614: Moved to utils

common/copy_kvp_client.sh
abhamidipati0614: Changed the dest dir

jithinjosepkl (approved):

PR #243, opened 2023-05-17
add check to verify that yum is updated

LiquidPT:


tests/run-tests.sh
```diff
@@ -443,6 +443,14 @@ check_exit_code() {
     fi
 }
 
+# verify that yum is updated
+case ${distro} in
+    Ubuntu*) sudo apt-get -q --assume-no update;;
+    CentOS* | AlmaLinux*) sudo yum update -n;;
```
abhamidipati0614, line 449: Pls use --assumeno
yum update --assumeno

abhamidipati0614, line 449: pls use --setopt tsflags=test flag
`yum update -y --setopt tsflags=test`
and then clean the cache using `yum clean packages`

tests/run-tests.sh
```diff
@@ -443,6 +443,14 @@ check_exit_code() {
     fi
 }
 
+# verify that yum is updated
```
abhamidipati0614, line 446: we are checking for updates on all the distros. pls change yum to package as "verify if package update works"

abhamidipati0614 (changes requested):

tests/run-tests.sh
```diff
@@ -443,6 +443,17 @@ check_exit_code() {
     fi
 }
 
+# verify that yum is correctly updated
```
abhamidipati0614, line 446: pls correct this... it not only pertains to yum but apt as well

tests/run-tests.sh
```diff
@@ -443,6 +443,17 @@ check_exit_code() {
     fi
 }
 
+# verify that yum is correctly updated 
+case ${distro} in
+    Ubuntu*) sudo apt-get -q --assume-no update;;
+    CentOS* | AlmaLinux*) 
+        yum update -y --setopt tsflags=test
```
abhamidipati0614, line 450: semicolons at the end of each line inside a case condition

abhamidipati0614 (changes requested):

tests/run-tests.sh
abhamidipati0614: Oh! you also need to run it as sudo

LiquidPT: > semicolons at the end of each line inside a case condition

I think you only need the double semicolons at the end of the condition. No other semicolons at the end of each line withing the condition. That's how I found it in the examples.

abhamidipati0614 (approved):

PR #252, opened 2023-05-30
Add slurm in Ubuntu images

LiquidPT:


ubuntu/common/install_utils.sh
```diff
@@ -43,7 +43,8 @@ apt-get -y install numactl \
                    libkeyutils1 \
                    libkeyutils-dev \
                    libmount-dev \
-                   nfs-common
+                   nfs-common \
+                   slurm-wlm
```
abhamidipati0614, line 47: can you pls also add slurm to alma scripts?

tests/run-tests.sh
```diff
@@ -619,6 +623,12 @@ then
     module unload mpi/hpcx
 fi
 
+# verify slurm is installed
+if [ $CHECK_SLURM -eq 1 ]
+then
+    check_exists "/usr/sbin/slurmd-wlm"
```
abhamidipati0614, line 629: it is good to check slurm version and also if the service is active

PR #255, opened 2023-06-01
Moneo update

rafsalas19:
- Update Moneo install
- install pssh


abhamidipati0614 (approved):
Please verify if yum update works

PR #262, opened 2023-06-20
SUSE config hardend

petersatsuse:
Some changes to make the setup more stable and secure
Some changes to make the dependency tree smaller


suse/sle-hpc-15.x/common/install_nvidiagpudriver.sh
```diff
@@ -25,14 +25,19 @@ CUDA_DASH_VERSION=${CUDA_VERSION/./-}
...
+zypper -n install -y -l --no-recommends cuda-toolkit-${CUDA_DASH_VERSION} cuda-compiler-${CUDA_DASH_VERSION} cuda-command-line-tools-${CUDA_DASH_VERSION} gds-tools-${CUDA_DASH_VERSION} cuda-libraries-${CUDA_DASH_VERSION}  nvidia-fabricmanager = ${NVIDIA_VERSION} "nvidia-gfxG05-kmp-azure = ${NVIDIA_VERSION}" "nvidia-computeG05 = ${NVIDIA_VERSION}"
+
 
 $COMMON_DIR/write_component_version.sh "CUDA" ${CUDA_VERSION}
 $COMMON_DIR/write_component_version.sh "NVIDIA" ${NVIDIA_VERSION}
 
 # Post-install tasks (version its set through 'alternatives')
 echo 'export PATH=$PATH:/usr/local/cuda/bin' | tee -a /etc/bash.bashrc.local
-echo 'export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda/lib64' | tee -a /etc/bash.bashrc.local
+echo '/usr/local/cuda/lib64' | tee /etc/ld.so.conf.d/cuda.conf
```
abhamidipati0614, line 40: Any specific reason for not writing CUDA library path to bashrc and only writing it to CUDA conf?

suse/sle-hpc-15.x/sle-hpc-15-sp4/fixes.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
abhamidipati0614, line 1: How is this file different from install_utils.sh. Is there any reason why this is named as fixes.sh?

suse/sle-hpc-15.x/common/install_nvidiagpudriver.sh
petersatsuse: There are several problems with LD_LIBRARY_PATH...

-  It compromises the security of a system, as someone can access your application to load a shared library with malicious code as LD_LIBRARY_PATH overwrites the normal search path.

- It decreases the performance of your applications if the path has a lot of directories because failed calls will increase.

- LD_LIBRARY_PATH will cause an application to load a shared library that’s not linked against which might not be compatible with the original version.

So there is no need to set LD_LIBRARY_PATH, if the cuda libraries are written to the global ld.so.conf. This uses the dynamic linker to figure out which path to search for the library file at runtime (which is the default behaviour). 

Second, using ld.so.conf works for all users and programs without doing explicit settings for every user in bashrc or similar.

LD_LIBRARY_PATH is useful for things that are temporary and for user based libraries installed in the user's home directory, as normal users do not have access to /etc/ld.so.conf, but should not be the default for a system and not used in a global context.

suse/sle-hpc-15.x/sle-hpc-15-sp4/fixes.sh
petersatsuse: Fixes are changes which are required during the lifecycle of the version.
As install_utils is move to common as its should be valid for all e.g. 15 versions we would need something in addition for a case.
Yes, it would be not necessary if install_utils would stay in the minior version tree

abhamidipati0614: I'd prefer the install_utils.sh inside of 15sp4 tree, instead of 'fixes', to be in sync with the rest of the repo

abhamidipati0614 (approved):

PR #270, opened 2023-08-02
Amd mi200: Update ROCm to 5.6

hpourreza:


hpourreza:
Not sure why it is showing 3 commits instead of the one that updated ROCm from 5.4 to 5.6

PR #287, opened 2023-10-11
Move versioning for RDMA Core into config

LiquidPT:


requirements.json
```diff
@@ -215,5 +215,13 @@
         "ubuntu22.04": {
             "version": "0.2.4"
         }
+    },
+    "rdma_core": {
+        "almalinux8.7": {
+            "branch": "stable-v34"
+        },
+        "ubuntu22.04": {
+            "branch": "stable-v34"
```
abhamidipati-msft, line 224: I think we can update this to the latest one

PR #292, opened 2023-10-30
Update environment-modules version for Alma

lorisercole:
The version of Environment Modules that [comes with AlmaLinux 8](https://repology.org/project/environment-modules/versions) is quite outdated (4.5.2).

We would like to improve the user experience with modules, in particular by showing only the important ones while hiding all the unnecessary dependencies, thanks to a functionality introduced in [version 4.7](https://modules.readthedocs.io/en/latest/MIGRATING.html#v4-7), that can be exploited by spack (see [this PR](https://github.com/spack/spack/pull/36619)).
I therefore propose to install an updated version of `environment-modules`.
Here I use version 5.0.1, which is the one shipped with AlmaLinux 9.


alma/alma-8.x/common/install_utils.sh
```diff
@@ -53,6 +52,11 @@ yum install -y numactl \
     gcc-gfortran \
     perl
 
+# Install environment-modules 5.0.1
+wget https://repo.almalinux.org/vault/9.1/BaseOS/x86_64/os/Packages/environment-modules-5.0.1-1.el9.x86_64.rpm
```
abhamidipati-msft, line 56: Is the compatibility tested? We currently only support AlmaLinux HPC 8.7 and the vault only shows support for 4.5.2

lorisercole: I tried to install it and run it on AlmaLinux HPC 8.7 and I did not encounter any issues

jithinjosepkl (approved):

ltalirz:
Given this is approved, anything stopping this from being merged?

PR #299, opened 2023-11-30
Removing images that won't be converted to spack

LiquidPT:
When we merge the spack codebase into master, we will be archiving the older images and won't be maintaining them side-by-side with the spack code


abhamidipati-msft:
Thank you @LiquidPT
Can you verify by running the pipeline for suitable distros?

PR #300, opened 2023-12-14
Fix Alma 8.7 install in Spack

LiquidPT:
Alma 8.7 was failing to install using Spack. While fixing this, I made a couple of small quality of life changes as well


abhamidipati-msft (approved):

alma/common/install_spack.sh
```diff
@@ -41,4 +41,5 @@ $COMMON_DIR/write_component_version.sh "spack" $spack_version
 # Create an environment/ container in /opt
 spack env create -d $HPC_ENV
 echo "spack env activate $HPC_ENV" | tee -a /etc/profile
+echo "PATH=\$(echo \"\$PATH\" | tr \":\" \"\\n\" | grep -v \"$HPC_ENV/.spack-env/view/bin\" | tr \"\\n\" \":\" | sed \"s/:$//\")" | sudo tee -a /etc/profile
```
abhamidipati-msft, line 44: Maybe I am missing some context...did we have to add spack view also to the PATH?

LiquidPT: Honestly, I don't remember why this is here. We had to add it to Ubuntu to get it working, so I copied it over.

PR #314, opened 2024-02-14
fix swapped values in ncv4 topo file

shivanispatel:
The GPU values for NUMA domains 1 and 2, 3 and 4, were swapped. Corrected the values and "unswapped" them. 

Ran NCCL and verified that there were no errors related to the change in topo file. Attached the results as a .txt file.
[ncv4-topo-swap-test.txt](https://github.com/Azure/azhpc-images/files/14289622/ncv4-topo-swap-test.txt)


rafsalas19 (approved):
Everything looks good. For clarity, Shivani shared the results of lstopo and they now match the fixed topo file. Additionally, she shared NCCL results with debug info and no topo issues were indicated.

PR #324, opened 2024-03-08
Code cleanup and performance improvements

LiquidPT:
* Use the spack binary cache to download common packages recompiled
* Consolidate some installation code to common files
* Move paths added to the beginning for priority
* Upgrade GCC to a version commonly used in the spack binary cache, and enable a unit test to insure it's installed

PR #325, opened 2024-03-13
Code cleanup and performance improvements (#324)

sangeetha-tudum2209:
* Use the spack binary cache to download common packages recompiled
* Consolidate some installation code to common files
* Move paths added to the beginning for priority
* Upgrade GCC to a version commonly used in the spack binary cache, and enable a unit test to insure it's installed


LiquidPT (changes requested):
The spack branch is set to be deleted. Use main.

PR #326, opened 2024-03-14
Upgrade all components 

LiquidPT:
* In Ubuntu, upgrade all components (except kernel) prior to install of HPC components
* Remove slurm install for now


LiquidPT:
Checking in as per chat confirmation from @jithinjosepkl

PR #330, opened 2024-04-05
copy hpcx tests to hpcx-rebuild

LiquidPT:
The OSU Benchmarks were not in the directory where HPC-X was rebuilt with PMIx. Copied the test directory and adjusted the unit tests to catch this case.


jeseszhang1010 (approved):
approved. 
Curious why we rebuild hpcx with pmix, maybe I missed that context.

LiquidPT:
> approved. Curious why we rebuild hpcx with pmix, maybe I missed that context.

We had a request for pmix and needed to use the version from cyclecloud for compatibility. Hpcx comes prebuilt, so we need to rebuild it using this external pmix library

jeseszhang1010:
> > approved. Curious why we rebuild hpcx with pmix, maybe I missed that context.
> 
> We had a request for pmix and needed to use the version from cyclecloud for compatibility. Hpcx comes prebuilt, so we need to rebuild it using this external pmix library

Okay, CC has their special requirement on pmix. They'd better ensure it's working with openmpi (hpcx has in fact), as openmpi typically ingests particular pmix as submodule IIRC.

PR #338, opened 2024-04-24
Update to AzNHC install

rafsalas19:
- remove old install method
- pull docker image onto image

### DIFF
--- a/common/install_health_checks.sh
+++ b/common/install_health_checks.sh
@@ -1,11 +1,11 @@
 #!/bin/bash
+
 set -e
 
-# Set the azhc version
-AZHC_VERSION=$(jq -r '.azhc."'"$DISTRIBUTION"'".version' <<< $COMPONENT_VERSIONS)
+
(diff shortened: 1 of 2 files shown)


common/install_health_checks.sh
```diff
@@ -1,22 +1,22 @@
 #!/bin/bash
+
+
 set -e
 
-# Set the azhc version
-AZHC_VERSION=$(jq -r '.azhc."'"$DISTRIBUTION"'".version' <<< $COMPONENT_VERSIONS)
+AZHC_VERSION=v0.4.2
```
LiquidPT, line 6: We should be using the version in requirements.json. No more hardcoding versions in files

LiquidPT (changes requested):

common/install_health_checks.sh
rafsalas19: sounds good let me modify

rafsalas19: addressed

LiquidPT (approved):

Issue #327, opened 2024-03-15
NCCL graph and topology incompatible with A100


r-b-g-b:
I'm using the ubuntu-hpc 2204 x64 Gen 2 image on a Standard NC24ads A100 v4 VM.

I train a vLLM model that uses NCCL and observe the following error:

<details>
  <summary>Error</summary>

::16674:16674 [0] NCCL INFO Bootstrap : Using eth0:10.1.0.4<0>
::16674:16674 [0] NCCL INFO NET/Plugin : Plugin load (libnccl-net.so) returned 2 : libnccl-net.so: cannot open shared object file: No such file or directory
::16674:16674 [0] NCCL INFO NET/Plugin : No plugin found, using internal implementation
::16674:16674 [0] NCCL INFO cudaDriverVersion 12020
NCCL version 2.18.6+cuda11.8
::16674:17023 [0] NCCL INFO NET/IB : Using [0]mlx5_an0:1/RoCE [RO]; OOB eth0:10.1.0.4<0>
::16674:17023 [0] NCCL INFO Using network IB
::16674:17023 [0] NCCL INFO comm 0x5640f24752d0 rank 0 nranks 1 cudaDev 0 nvmlDev 0 busId 100000 commId 0x2064c00fc2f91516 - Init START
::16674:17023 [0] NCCL INFO NCCL_TOPO_FILE set by environment to /opt/microsoft/ncv4/topo.xml
::16674:17023 [0] NCCL INFO Setting affinity for GPU 0 to ffffff
::16674:17023 [0] NCCL INFO NCCL_GRAPH_FILE set by environment to /opt/microsoft/ncv4/graph.xml

::16674:17023 [0] graph/search.cc:703 NCCL WARN XML Import Channel : dev 1 not found.
::16674:17023 [0] NCCL INFO graph/search.cc:733 -> 2
::16674:17023 [0] NCCL INFO graph/search.cc:740 -> 2
::16674:17023 [0] NCCL INFO graph/search.cc:840 -> 2
::16674:17023 [0] NCCL INFO init.cc:880 -> 2
::16674:17023 [0] NCCL INFO init.cc:1358 -> 2
::16674:17023 [0] NCCL INFO group.cc:65 -> 2 [Async thread]
::16674:16674 [0] NCCL INFO group.cc:406 -> 2
::16674:16674 [0] NCCL INFO group.cc:96 -> 2
Traceback (most recent call last):

...

[...]

PR #347, opened 2024-05-22
Update SKU Customizations for NCv4

jithinjosepkl:
NCv4 topo file and graph file is applicable only for the NCv4 96-core VM size. Smaller sizes dont need topo/graph file.


jithinjosepkl:
Fixes #327

LiquidPT (approved):

PR #348, opened 2024-05-31
add cleanup_esm.sh to remove ubuntu esm packages

vgamayunov:
remove graphviz package (+deps) requiring Ubuntu Pro for security updates

### DIFF
--- a/ubuntu/common/cleanup_esm.sh
+++ b/ubuntu/common/cleanup_esm.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# Remove packages requiring Ubuntu Pro for security updates
+apt-get update
+apt-get remove --auto-remove -y graphviz
--- a/ubuntu/common/cleanup_insecure_packages.sh
+++ b/ubuntu/common/cleanup_insecure_packages.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-# Remove packages requiring Ubuntu Pro for security updates
-apt-get update
-apt-get remove --auto-remove -y graphviz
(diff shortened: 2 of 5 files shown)


darkwhite29:
I would say maybe using a more readable name for the newly added script, like: clean_insecure_package.sh

LiquidPT:
> I would say maybe using a more readable name for the newly added script, like: cleanup_insecure_package.sh

I'd actually prefer to see package removal added to `install_prerequisites.sh` (and move it to /ubuntu/common rather than having version specific ones since they are the same).

This would mean that base image package cleanup could be early (and where we're already doing an update/upgrade) and allow for other things installed later that might be dependent on it to catch that it's missing (not that I think that will happen in this case)

ubuntu/common/cleanup_esm.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
LiquidPT, line 1: duplicate file

ubuntu/common/cleanup_insecure_packages.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Remove packages requiring Ubuntu Pro for security updates
+apt-get update
```
LiquidPT, line 4: suggest a more general name since this might be used for other purposes later. `cleanup_unused_packages.sh` or `remove_unused_packages.sh`

Also, I'm not sure we need to do another update here. It happens in `install_prerequisites.sh` a few lines before. I suppose it doesn't hurt, but seems redundant.

vgamayunov:
agree with the comments - updated

LiquidPT (approved):

PR #354, opened 2024-06-14
remove references to the azhpcstor storage account

LiquidPT:


alma/alma-8.x/common/install_utils.sh
```diff
@@ -77,10 +77,6 @@ wget -r --no-parent -A "subunit-*.el8.x86_64.rpm" https://dl.fedoraproject.org/p
 yum localinstall ./dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/s/subunit-[0-9].*.el8.x86_64.rpm -y
 yum localinstall ./dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/s/subunit-devel-[0-9].*.el8.x86_64.rpm -y
 
-# Download jq utility
-wget -r --no-parent -A "jq-*.el8.x86_64.rpm" https://repo.almalinux.org/almalinux/8/AppStream/x86_64/os/Packages/
-yum localinstall ./repo.almalinux.org/almalinux/8/AppStream/x86_64/os/Packages/jq-*.el8.x86_64.rpm -y
-
```
jeseszhang1010, line 83: Is it built-in somewhere, so we don't need to download/install it any more?

LiquidPT: We install it in `install_prerequisites.sh` so it's already installed by the time we get here.

alma/alma-8.x/common/install_amd_libs.sh
```diff
@@ -10,13 +10,15 @@ AOCL_VERSION=$(jq -r '.aocl.version' <<< $amd_metadata)
 AOCL_SHA256=$(jq -r '.aocl.sha256' <<< $amd_metadata)
 
 TARBALL="aocl-linux-aocc-${AOCL_VERSION}.tar.gz"
-AOCL_DOWNLOAD_URL=https://azhpcstor.blob.core.windows.net/azhpc-images-store/${TARBALL}
+BASE_AOCL_VERSION=${AOCL_VERSION:0:3}
```
yakovdyadkin, line 13: Is this just grabbing the major.minor version?

LiquidPT: Yes. In the case of 4.2.0, it grabs 4.2 (this would break down if it were 4.12, but I haven't had any evidence of that. we'll go down that road if we get there)

yakovdyadkin: Yeah, that's what I was leaning towards. Leaving this for later, but this should be good for up to a three parted version string
`echo $VERSION | grep -Eo '[0-9]+\.[0-9]+'`

PR #358, opened 2024-07-10
Adding ability to choose ROCm or CUDA SKUs

moes1:
A couple changes: 

1. Added "SKU" variable that should be ROCm or CUDA, so we can switch between installing ROCm or CUDA libraries. It is always CUDA except for Ubuntu 22.04 where it must be chosen. 
2. Added tools needed to install ROCm libraries to create a ROCm VM.


ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rocm.sh
```diff
@@ -0,0 +1,71 @@
...
+cat /etc/security/limits.conf | sed -e "s/$line/$string_ha/" > temp_file.txt
+sudo mv temp_file.txt /etc/security/limits.conf
+
+cat /etc/security/limits.conf | grep -v "  stack" > tmplimits.conf
+sudo mv tmplimits.conf /etc/security/limits.conf
+
+echo blacklist amdgpu | sudo tee -a /etc/modprobe.d/blacklist.conf
+sudo update-initramfs -c -k $(uname -r)
+
+echo "Writing gpu mode probe in init.d"
```
yakovdyadkin, line 33: Can you use a heredoc instead of individual echo cmds?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+set -ex
+
+#sudo apt-get install -y rocblas rccl-dev rccl-rdma-sharp-plugins
```
yakovdyadkin, line 4: Is this line still needed? And does L6 have any relation to it?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
yakovdyadkin, line 1: More style related, but can you use `pushd` and `popd` instead of `cd ~` to be consistent with the other installation scripts.

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -ex
 
+export GPU=$1
```
LiquidPT, line 4: Should we validate the incoming param? As it is, I could send in an invalid param (even "nvidia" in lowercase) and not know for 45 min as the installation script would proceed as normal and skip over all of the GPU stuff.

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -0,0 +1,59 @@
...
+
+
+sudo sysctl kernel.numa_balancing=0
+echo "kernel.numa_balancing=0" | sudo tee -a /etc/sysctl.conf
+
+
+git clone https://github.com/ROCmSoftwarePlatform/rccl-tests
+cd rccl-tests
+
+source /opt/hpcx-v2.16-gcc-mlnx_ofed-ubuntu22.04-cuda12-gdrcopy2-nccl2.18-x86_64/hpcx-init.sh
```
yakovdyadkin, line 25: Is it possible to use the hpcx environment module at this point, and use some of the environment variables once the module instead of hardcoding HPCX var?

customizations/ndv5_rocm.sh
```diff
@@ -0,0 +1,22 @@
...
+# Link the NDv5 topology file into /opt/microsoft/ndv5/
+ln -sf /opt/microsoft/ndv5-topo.xml /opt/microsoft/ndv5/topo.xml
+
+## Set NCCL configuration file for NDv5
+bash -c "cat > /etc/nccl.conf" <<'EOF'
+NCCL_IB_PCI_RELAXED_ORDERING=1
+NCCL_TOPO_FILE=/opt/microsoft/ndv5/topo.xml
+EOF
+
+error_code=$?
```
LiquidPT, line 16: Fabric manager isn't installed, and there's no line here to start it, so no need for the error code check

customizations/ndv4_rocm.sh
```diff
@@ -0,0 +1,38 @@
...
+EOF
+
+error_code=$?
+if [ ${error_code} -ne 0 ]
+then
+    echo "NVIDIA Fabic Manager Inactive!"
+    exit ${error_code}
+fi
+
+# ## Setup NVME devices
```
LiquidPT, line 23: remove commented out lines?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rocm.sh
moes1: done

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rocm.sh
```diff
@@ -0,0 +1,71 @@
...
+echo -e '[Service]\n\nExecStart=/etc/init.d/initamdgpu.sh\n\n' \
+               | sudo tee -a rocmstartup.service
+echo -e '[Install]\n\nWantedBy=multi-user.target' \
+               | sudo tee -a rocmstartup.service
+
+sudo mv rocmstartup.service /etc/systemd/system/rocmstartup.service
+sudo systemctl start rocmstartup
+sudo systemctl enable rocmstartup
+
+sudo apt install rocm-bandwidth-test
```
yakovdyadkin, line 71: Is this non interactive or is the `-y` arg missing?

common/setup_sku_customizations.sh
```diff
@@ -73,6 +77,67 @@ rm -rf /opt/microsoft/ndv5
...
+    *) echo "No SKU customization for \$vmSize";;
+esac
+EOF
+chmod 755 /usr/sbin/setup_sku_customizations.sh
+
+## Systemd service for removing SKU based customizations
+cat <<EOF >/usr/sbin/remove_sku_customizations.sh
+#!/bin/bash
+
+# Stop nvme raid service
```
LiquidPT, line 119: remove commented out lines?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+set -ex
+
+#sudo apt-get install -y rocblas rccl-dev rccl-rdma-sharp-plugins
+sudo apt install libstdc++-12-dev
+sudo apt remove -y rccl
```
LiquidPT, line 6: These installation scripts must be run as sudo, so the sudo can be removed from all lines here

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rocm.sh
LiquidPT: I ran into this problem recently where I left that out, and it caused the script to timeout waiting (and didn't run everything after it)

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
LiquidPT: Especially since that will change with a new version of HPCX

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install.sh
LiquidPT, line 4: Should update the readme to document this

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
moes1: How about I source /opt/hpcx*/hpcx-init.sh ?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
moes1: removed. I do want to uninstall rccl to keep only one version of rccl installed.

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
moes1: addressed

alma/alma-8.x/alma-8.7-hpc/install.sh
```diff
@@ -4,6 +4,8 @@ set -ex
 # install pre-requisites
 ./install_prerequisites.sh
 
+export GPU="NVIDIA"
```
LiquidPT, line 7: I'd like to add the param to the other install scripts as well. Break the interface once and make it consistent.

If we want to only support AMD on U22 for now, that's fine. But we should take in the param, do validation similar to what we add to U22, but return an error message that "AMD GPUs are not supported on this distro at this time" or something similar.

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rocm.sh
moes1: Ah I was not hitting a request to prompt because rocm-bandwidth-test doesn't install any other dependencies but I will add -y in case it does at some later time.

alma/alma-8.x/alma-8.7-hpc/install.sh
moes1: You can pass an argument to these install scripts even though it isn't read if that makes the pipeline simpler to implement. But you are saying you want it to fail loudly if someone tries to generate an AMD image for Ubuntu 20 and Alma?

customizations/ndv4_rocm.sh
moes1: these are present in ndv4.sh as well, should we remove them there?

common/setup_sku_customizations.sh
moes1: done

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
moes1: done

customizations/ndv4_rocm.sh
moes1: removed in both files.

customizations/ndv5_rocm.sh
moes1: removed

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install.sh
moes1: I've added a check for this. By default GPU will be set to NVIDIA unless "AMD" is passed in, in which case it will be set to AMD.

alma/alma-8.x/alma-8.7-hpc/install.sh
moes1: I've added checks so that alma and ubuntu-20 install scripts will fail if anything other than NVIDIA is passed to them, but by default they will build the NVIDIA image if nothing is passed in.

LiquidPT (dismissed):

LiquidPT (approved):

PR #359, opened 2024-07-10
Clean up Ubuntu and Alma directories

LiquidPT:
This cleans up extra files within the Ubuntu and Alma directories, consolidating functionality within each of those as much as possible. 

Then next step will be to find commonalities between the distros that can be moved to /common

I do not know what the changes to the 3 MD files in root are. When I changed laptops this month, it would not let me discard changes, nor could it show me any visible changes. The comparison here also doesn't show any changes.


alma/common/hpc-tuning.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 
 # Disable some unneeded services by default (administrators can re-enable if desired)
+systemctl disable wpa_supplicant
```
yakovdyadkin, line 4: Just curious why the additional services need to be disabled?

yakovdyadkin:
Do you think the 3 MD file changes are related to line endings, considering line endings are now explicitly set to LF in gitattributes?

yakovdyadkin (approved):

alma/common/hpc-tuning.sh
LiquidPT: > Just curious why the additional services need to be disabled?

this was one of my questions for you about the differences in the existing files. I took this from the existing alma 8.x hpc-tuning.sh

LiquidPT:
> Do you think the 3 MD file changes are related to line endings, considering line endings are now explicitly set to LF in gitattributes?

Maybe, though I added tried removing the .gitattributes file and it should let me undo it as well. If it is line0endings, it's probably good to standardize on linux-style anyways.

LiquidPT:
> Do you think the 3 MD file changes are related to line endings, considering line endings are now explicitly set to LF in gitattributes?

Maybe, though I added tried removing the .gitattributes file and it should let me undo it as well. If it is line0endings, it's probably good to standardize on linux-style anyways.

LiquidPT:
> Do you think the 3 MD file changes are related to line endings, considering line endings are now explicitly set to LF in gitattributes?

Maybe, though I added tried removing the .gitattributes file and it should let me undo it as well. If it is line0endings, it's probably good to standardize on linux-style anyways.

PR #361, opened 2024-07-16
Added Torset Tool

sandeeponcall:


tools/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Prepare a venv:
+
+```bash
+python3 -m venv torset-env
+source torset-env/bin/activate
+```
+
+```bash
+pip install -r requirements. txt
```
LiquidPT, line 15: There's an extra space before `txt`

sandeeponcall: done, fixed the extra space

tools/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```bash
+ python3 torset-tool.py --hosts <path-to-host-file> --pkey_path  <path-to-ssh-private-key>  --sharp_cmd_path <path-to-sharp-cmd> --output_dir  <path-to-output-dir>
+```
+
+### Outputs
+
+This will create a number of files in the <output> directory:
+
+- guids.txt: A file with the InfiniBand device GUIDs from every host
```
moes1, line 28: Should we clarify it is the port GUIDs?

moes1:
The plan is just to put all tools directly in the "tools" directory and update this one readme file?

LiquidPT:
> The plan is just to put all tools directly in the "tools" directory and update this one readme file? And did we want to install this in the image somewhere @jithinjosepkl ?

This was also my question. It's a `/tools` directory, but the README top level heading is the Torset tool. 

Suggest subdirectories, stand-alone MD files (that the README points to), or Torset being a subheading in the README (though this could get unwieldy depending on how many tools occur)

jithinjosepkl:
> The plan is just to put all tools directly in the "tools" directory and update this one readme file? And did we want to install this in the image somewhere @jithinjosepkl ?

Yes, this should be part of the image.

LiquidPT (approved):

PR #362, opened 2024-07-17
add --bind-to numa to mpirun

shivanispatel:
Add `--bind-to numa` argument to mpirun commands.


jeseszhang1010 (approved):
LGTM. Please add why we have such change into commit message

LiquidPT (approved):

PR #363, opened 2024-07-25
FM-test

jeseszhang1010:
This PR adds a workaround to make FM service start automatically on top of branch matt/liquidpt/fm-test.
This is ONLY to share the workaround with team


LiquidPT:
I'll test this once I have the docker install fixed.

PR #364, opened 2024-07-25
Remove disabling containerd plugins

LiquidPT:


jeseszhang1010 (approved):
Good to me, approved 
Just curious.
What was the reason we had these disablement earlier?
Are these containerd plugins disablement break something now?

Could you add some high level description in the commit message?

LiquidPT:
These plug-ins were disabled in response to error messages in log files (but container still started). It seems that the config modifications were made incorrectly as containerd won't start. Commenting the problematic line for further testing.

PR #365, opened 2024-07-25
Upgrade to NVIDIA driver and Fabric Manager 550

LiquidPT:
Updating NVIDIA driver and Fabric Manager to 550.90

Previously we were having issues with Fabric Manager starting on VM restart, so this includes a workaround for this issue


jeseszhang1010 (approved):

jeseszhang1010 (approved):
>update nvidia driver and FM to 550.45.15
Minor typo, should be 550.54.15 :)

LiquidPT:
> > update nvidia driver and FM to 550.45.15
> 
> Minor typo, should be 550.54.15 :)

That was the original iteration of 550 we installed. We're installing 550.90.07 now.

PR #366, opened 2024-07-25
July 2024 component updates

LiquidPT:
* updated components to latest available
* updated code to work with new components (mostly paths)
* cleaned up some of the versioning in the `versions.json` to be more consistent


versions.json
```diff
@@ -236,12 +219,12 @@
...
     "rdma_core": {
         "common": {
-            "branch": "stable-v48"
+            "branch": "v52.0"
         }
     },
     "pmix": {
         "common": {
-            "version": "4.2.9-1"
+            "version": "4.2.9"
```
yakovdyadkin, line 227: Why is the release omitted?

versions.json
```diff
@@ -236,12 +219,12 @@
     },
     "rdma_core": {
         "common": {
-            "branch": "stable-v48"
+            "branch": "v52.0"
```
yakovdyadkin, line 222: Just going with the latest version, not necessarily stable (as compared to before)?

versions.json
```diff
@@ -100,88 +100,88 @@
...
+            "version": "2024.2.0.664",
+            "sha256": "f1f46f5352c197a9840e08fc191a879dad79ebf742fe782e386ba8006f262f7a",
+            "url": "https://registrationcenter-download.intel.com/akdlm/IRC_NAS/cdff21a5-6ac7-4b41-a7ec-351b5f9ce8fd/l_onemkl_p_2024.2.0.664_offline.sh"
         }
     },
     "waagent": {
         "ubuntu20.04": {
-            "version": "2.9.0.4",
-            "sha256": "040969f507f73f3a2c95d5b0568225ad68f7f91bfec99bd92154c3fa9e28034b"
+            "version": "2.11.1.4",
```
yakovdyadkin, line 179: Possible to use a common version for waagent?

versions.json
```diff
@@ -100,88 +100,88 @@
...
         "ubuntu20.04": {
             "driver": {
-                "version": "12-4",
+                "version": "12.5",
                 "distribution": "ubuntu2004"
             },
             "samples": {
-                "version": "12.4",
-                "sha256": "aa28fa2227768dd31ebbf9cd48b265a0c8810fae03e02c6079c0fa71bbea7319"
+                "version": "12.5",
```
yakovdyadkin, line 115: Would it be possible to move samples under a common distribution?

versions.json
```diff
@@ -1,19 +1,19 @@
 {
     "azcopy": {
```
yakovdyadkin, line 2: Possible to move everything under common distro?

alma/common/install_nvidia_fabric_manager.sh
```diff
@@ -10,8 +10,10 @@ NVIDIA_FABRICMANAGER_DISTRIBUTION=$(jq -r '.distribution' <<< $nvidia_fabricmana
 NVIDIA_FABRICMANAGER_VERSION=$(jq -r '.version' <<< $nvidia_fabricmanager_metadata)
 NVIDIA_FABRICMANAGER_SHA256=$(jq -r '.sha256' <<< $nvidia_fabricmanager_metadata)
 
-NVIDIA_FABRIC_MNGR_URL=http://developer.download.nvidia.com/compute/cuda/repos/${NVIDIA_FABRICMANAGER_DISTRIBUTION}/x86_64/nvidia-fabric-manager-${NVIDIA_FABRICMANAGER_VERSION}.x86_64.rpm
+NVIDIA_FABRIC_MNGR_URL=http://developer.download.nvidia.com/compute/cuda/repos/${NVIDIA_FABRICMANAGER_DISTRIBUTION}/x86_64/nvidia-fabric-manager-${NVIDIA_FABRICMANAGER_VERSION}-1.x86_64.rpm
+FILENAME=$(basename $NVIDIA_FABRIC_MNGR_URL)
```
yakovdyadkin, line 14: Nitpick, can this be named `NVIDIA_FABRIC_MNGR_RPM` or something more descriptive?

common/clear_history.sh
```diff
@@ -65,6 +65,7 @@ rm -rf /tmp/nvidia* /tmp/MLNX* /tmp/ofed.conf /tmp/dkms* /tmp/*mlnx*
 rm -rf /run/cloud-init
 rm -rf /root/*
 rm -rf /usr/tmp/dnf*
+rm -f /opt/azurehpc/test/azurehpc-health-checks/health.log
```
yakovdyadkin, line 68: Would this break the build if NHC is not run during customization?

common/install_azcopy.sh
```diff
@@ -9,7 +9,7 @@ azcopy_version=$(jq -r '.version' <<< $azcopy_metadata)
 azcopy_release=$(jq -r '.release' <<< $azcopy_metadata)
 azcopy_sha256=$(jq -r '.sha256' <<< $azcopy_metadata)
 TARBALL="azcopy_linux_amd64_$azcopy_version.tar.gz"
-AZCOPY_DOWNLOAD_URL="https://azcopyvnext.azureedge.net/$azcopy_release/$TARBALL"
+AZCOPY_DOWNLOAD_URL="https://azcopyvnext.azureedge.net/releases/release-${azcopy_release}/${TARBALL}"
```
yakovdyadkin, line 12: Did the path change for newer releases of azcopy?

ubuntu/common/install_nvidia_fabric_manager.sh
```diff
@@ -10,9 +10,11 @@ NVIDIA_FABRICMANAGER_DISTRIBUTION=$(jq -r '.distribution' <<< $nvidia_fabricmana
 NVIDIA_FABRICMANAGER_VERSION=$(jq -r '.version' <<< $nvidia_fabricmanager_metadata)
 NVIDIA_FABRICMANAGER_SHA256=$(jq -r '.sha256' <<< $nvidia_fabricmanager_metadata)
 
-NVIDIA_FABRICMANAGER_PREFIX=$(echo $NVIDIA_FABRICMANAGER_VERSION | cut -d '_' -f1)
-NVIDIA_FABRIC_MNGR_URL=http://developer.download.nvidia.com/compute/cuda/repos/${NVIDIA_FABRICMANAGER_DISTRIBUTION}/x86_64/nvidia-fabricmanager-${NVIDIA_FABRICMANAGER_VERSION}_amd64.deb
+NVIDIA_FABRICMANAGER_PREFIX=$(echo $NVIDIA_FABRICMANAGER_VERSION | cut -d '.' -f1)
+NVIDIA_FABRIC_MNGR_URL=http://developer.download.nvidia.com/compute/cuda/repos/${NVIDIA_FABRICMANAGER_DISTRIBUTION}/x86_64/nvidia-fabricmanager-${NVIDIA_FABRICMANAGER_PREFIX}_${NVIDIA_FABRICMANAGER_VERSION}-1_amd64.deb
+FILENAME=$(basename $NVIDIA_FABRIC_MNGR_URL)
```
yakovdyadkin, line 15: Same nitpick comment as the for alma's nvidia fabric manager installation.

yakovdyadkin (approved):

versions.json
LiquidPT: was always `-1` on a lot of these. I can bring it back if it makes sense, but I was trying to align versioning with what is shown on the websites for ease. (the `-1` is moved to the code)

versions.json
LiquidPT: possible, but highlighting for the moment that alma DOESN'T use the version file. I'd like to see if there's a common install strategy we want to use for WAAGENT, but didn't want to press to do it in this PR

versions.json
LiquidPT: As it's structured at the moment, no, as common only works at the top level. I'd need to pull cuda-driver and cuda-samples out separately (which might make sense).

The other option is to pull `distribution` out of the config (as it doesn't change release to release) and put it in `set_properties.sh`. That would allow all of `cuda` to be common. But if I did that I'd probably like to apply that pattern a few places and would probably be better suited for a new PR.

versions.json
LiquidPT: alma is different. In theory I could move the 2 ubuntus under common and have alma as an override, but then what happens when we add another distro like mariner? 

for this pass I had decided to just move things that were common across all to common just to simplify things and then look at more complicated patterns (or possibly more complicated logic) later

common/clear_history.sh
LiquidPT: NHC is always run in our pipelines.

I'll have to double check but I thought removing something that wasn't there was a no-op

common/install_azcopy.sh
LiquidPT: yup. new path. Also, I decided to take the word "release" out of the version file

tests/test-definitions.sh
```diff
@@ -227,11 +227,7 @@ function verify_lustre_installation {
 
 function verify_gdrcopy_installation {
     # Verify GDRCopy package installation
-    case ${ID} in
-        ubuntu) dpkg -l | grep gdrcopy;;
-        almalinux) dnf list installed | grep gdrcopy;;
-        * ) ;;
-    esac
+    gdrcopy_sanity
```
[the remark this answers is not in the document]
LiquidPT: it's part of gdrcopy when it's installed. First one mentioned under "Tests" in their readme

versions.json
LiquidPT: using tags rather than branches (maybe should rename this properties) for cleaner versioning.

That said, I see they have a `-2` release and so I see the argument of putting it back in.

versions.json
yakovdyadkin: I am thinking that if the release (`-1`) is moved to the code, there's a possibility that 2 files would need to be updated if the version is changed. i.e., it could be that the release is not always `-1`.

common/clear_history.sh
yakovdyadkin: Ah, my point is that customers that use these scripts might not use NHC in their build workflows. Maybe we can keep this commented and uncomment in our build pipeline (similar to some of the other cleanup/deprovision code)?

versions.json
LiquidPT: leaving as "branch", as we have the flexibility to use either here and I think we've needed to

yakovdyadkin (approved):

PR #367, opened 2024-07-26
Add additional tests for component installation

LiquidPT:
Adding some tests for components we install that aren't being checked now.

### DIFF
--- a/tests/test-definitions.sh
+++ b/tests/test-definitions.sh
@@ -193,7 +193,7 @@ function verify_aocl_installation {
 
 function verify_aocc_installation {
     # verify AMD compiler installation
-    check_exists "/opt/AMD/aocc-compiler-/${VERSION_AOCC}"
+    check_exists "/opt/AMD/aocc-compiler-${VERSION_AOCC}/"
 }
 
 function verify_docker_installation {
--- a/tests/test-definitions.sh
+++ b/tests/test-definitions.sh
@@ -217,8 +217,8 @@ function verify_ipoib_status {
 function verify_lustre_installation {
     # Verify lustre client package installation
     case ${ID} in
-        ubuntu) dpkg -l | grep amlfs-lustre-client;;
-        almalinux) dnf list installed | grep amlfs-lustre-client;;
+        ubuntu) dpkg -l | grep lustre-client;;
+        almalinux) dnf list installed | grep lustre-client;;
         * ) ;;
     esac
     check_exit_code "Lustre Installed" "Lustre not installed!"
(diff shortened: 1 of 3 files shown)


tests/test-definitions.sh
```diff
@@ -99,9 +99,8 @@ function verify_cuda_installation {
     check_exit_code "NVIDIA Peer memory module is inserted" "NVIDIA Peer memory module is not inserted!"
 
     # Verify if CUDA is installed
-    # re-enable this after testing
```
yakovdyadkin, line 102: Why was this commented before?

LiquidPT: oh, I need to recomment that out. It works ion VMs when I test, but fails in the pipeline for some reason. says that nvcc isn't a known command

tests/test-definitions.sh
```diff
@@ -210,6 +215,36 @@ function verify_ipoib_status {
...
+        * ) ;;
+    esac
+    check_exit_code "GDRCopy Installed" "GDRCopy not installed!"
+}
+
+function verify_pssh_installation {
+    # Verify PSSH package installation
+    case ${ID} in
+        ubuntu) dpkg -l | grep pssh;;
+        almalinux) dnf list installed | grep pssh;;
```
yakovdyadkin, line 242: I have seen pssh appear under different name such as `parallel-ssh` is the package name parallel ssh `pssh` across all distros?

yakovdyadkin:
General comment, with only 3 unique distro + version combinations in _tests/test-matrix.json_ that are almost identical in content, does it make sense to move the common tests that each distro specifies to `verify_common_components` func in _tests/run-tests.sh_?

yakovdyadkin (commented):
General comment, with only 3 unique distro + version combinations in tests/test-matrix.json that are almost identical in content, does it make sense to move the common tests that each distro specifies to verify_common_components func in tests/run-tests.sh?

moes1 (approved):

tests/test-definitions.sh
LiquidPT: on both Alma and Ubuntu we install a package named "pssh". 

We seem to use a mix of parallel ssh packages on our validation pipeline and I don't know why.

LiquidPT:
> General comment, with only 3 unique distro + version combinations in tests/test-matrix.json that are almost identical in content, does it make sense to move the common tests that each distro specifies to verify_common_components func in tests/run-tests.sh?

Maybe. There were certainly things that were different that I've recently made the same. I think I'd like to get Mariner in the mix before I make the call of "these things are also always installed".

PR #370, opened 2024-07-30
add sku customization for h200

shivanispatel:


common/setup_sku_customizations.sh
```diff
@@ -49,6 +49,9 @@ case \$vmSize in
     standard_nd96is*_h100_v5)
         /opt/azurehpc/customizations/ndv5.sh;;
 
+    standard_nd96is*_h200_v5)
```
LiquidPT, line 52: @moes1 , should this also be added to the AMD section?

jeseszhang1010: Or combine with pervious h100 case with 
standard_nd96is*_h[1-2]00_v5?

jeseszhang1010: > @moes1 , should this also be added to the AMD section?

No, h200 is Nvidia GPU.

LiquidPT: > > @moes1 , should this also be added to the AMD section?
> 
> No, h200 is Nvidia GPU.

Yes, so is H100 and it's in the AMD section. Thus I wanted to ask @moes1 . He added NVIDIA skus to the AMD section because he was using them to build AMD images when he didn't have access to AMD skus

jeseszhang1010: Got it, thanks for clarification.

shivanispatel: Made the change from h200 to checking for h[1-2]00 instead 
I stopped nvidia-fabricmanager, made the change in setup_sku_customizations.sh, rebooted the node, and then confirmed that nvidia-fabricmanager was started (as is done in ndv5.sh, which should be called by setup_sku_customizations.sh for h200)

LiquidPT (approved):

PR #373, opened 2024-08-19
Update UCX related components and reference bug

LiquidPT:
- Update DOCA, HPC-X, MVAPICH2, and IMPI
- Fix Ubuntu OpenMPI installation to use HCOLL and UCX bundled with the HPC-X

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -18,36 +18,36 @@
     },
     "doca": {
         "ubuntu20.04": {
-            "version": "2.7.0",
-            "sha256": "40b2480ce89c9ea8718c3f671eb01155de677031e0d51c2457ee5968055de3c7",
-            "url": "https://www.mellanox.com/downloads/DOCA/DOCA_v2.7.0/host/doca-host_2.7.0-209000-24.04-ubuntu2004_amd64.deb"
+            "version": "2.8.0",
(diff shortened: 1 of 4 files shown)


versions.json
```diff
@@ -18,42 +18,42 @@
...
+            "version": "2.20",
+            "sha256": "ad8398622ff19152be5bbec3abff6e801f2ad48941cfea478f2953d098e69a8c",
+            "url": "https://content.mellanox.com/hpc/hpc-x/v2.20/hpcx-v2.20-gcc-mlnx_ofed-redhat8-cuda12-x86_64.tbz"
         }
     },
     "mvapich2": {
         "common": {
-            "version": "2.3.7-1",
-            "sha256": "fdd971cf36d6476d007b5d63d19414546ca8a2937b66886f24a1d9ca154634e4"
+            "version": "2.3.7-azure",
```
yakovdyadkin, line 55: I understand `2.3.7-azure` is better for readability, but it's not actually used in the installer file. Do you think switching to 2.3.x (what is actually used) will bring more confusion in the long term?

yakovdyadkin (approved):

versions.json
LiquidPT: 1. yes it would
2. This is what Jithin told me to use

PR #377, opened 2024-09-05
Add ifconfig check to tests

LiquidPT:
We've run into cases where ibstat shows the infiniband devices, but ifconfig does not. Since this is a bad configuration and will cause issues, we should test ifconfig as well

### DIFF
--- a/tests/test-definitions.sh
+++ b/tests/test-definitions.sh
@@ -43,7 +43,7 @@ function verify_ib_device_status {
     check_exit_code "IB device state: LinkUp" "IB link not up"
 
     # verify ifconfig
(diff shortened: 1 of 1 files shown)


yakovdyadkin (approved):
Is the goal to explicitly check whether the IB device is configured before and after a reboot or can the checks be combined into a single if statement?

LiquidPT:
> Is the goal to explicitly check whether the IB device is configured before and after a reboot or can the checks be combined into a single if statement?

Updated the test to be a single line with an "or" in the grep

yakovdyadkin (approved):

PR #379, opened 2024-09-06
Rollback WAAgent version

LiquidPT:
The newer WAAgent has a bug that doesn't allow IP address assignment to the IB interfaces. Also unified the WAAgent installation code across distros.

### DIFF
--- a/common/install_waagent.sh
+++ b/common/install_waagent.sh
@@ -40,10 +40,11 @@ update_waagent_conf "OS.MonitorDhcpClientRestartPeriod" "60"
 update_waagent_conf "Provisioning.MonitorHostName" "y"
 update_waagent_conf "Provisioning.MonitorHostNamePeriod" "60"
 
-systemctl daemon-reload
-systemctl restart walinuxagent
 
 rm -rf WALinuxAgent-${WAAGENT_VERSION}
 
 $COMMON_DIR/write_component_version.sh "WAAGENT" $(waagent --version | head -n 1 | awk -F' ' '{print $1}' | awk -F- '{print $2}')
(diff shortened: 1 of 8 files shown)


tests/test-definitions.sh
```diff
@@ -43,7 +43,7 @@ function verify_ib_device_status {
     check_exit_code "IB device state: LinkUp" "IB link not up"
 
     # verify ifconfig
-    ifconfig | grep "ib0:\|ibP"
+    ifconfig | grep "ib[0-9]:\|ibP"
```
yakovdyadkin, line 46: 0-9? I don't think there's a SKU with more than 8 IB devices

common/install_waagent.sh
```diff
@@ -15,10 +17,34 @@ function update_waagent_conf {
     fi
 }
 
+# Set waagent version and sha256
```
yakovdyadkin, line 20: Can you add a comment WALinuxAgent is necessary to finish configuring IPoIB?

yakovdyadkin (approved):

tests/test-definitions.sh
LiquidPT: It just made more sense to me to "allow all digits" rather than saying 0-7.  Though if there's a more succinct way to do this ('d'?) then I would use that. 

When I asked yesterday I was trying to determine if I should allow for multiple digits.

yakovdyadkin: Totally forgot that 0-9 is how to grep for all digits, I think this make sense.

yakovdyadkin (approved):

PR #381, opened 2024-09-16
Fix IP issues in AlmaLinux

LiquidPT:
IP configuration kept getting overridden on Alma, causing AzNHC to fail. This was working fine during build, but creating a new VM from the image would revert the configuration. Added config to the SKU Customization service to change it on boot.


customizations/alma_disable_cloudinit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -ex
+
+# TEMP add to fix changed behavior in AlmaLinux 8.7
```
rafsalas19, line 4: can we add that the behavior is that cloud-init keeps reverting the /etc/sysconfig/network-scripts/ifcfg-eth0

rafsalas19 (approved):

PR #389, opened 2024-10-25
Onboard Azure Linux to functional tests

darkwhite29:


LiquidPT (approved):
I think these changes to run against AL3 look ok. Only by running them will we know if this is the complete changeset needed.

PR #390, opened 2025-01-09
Enable full Ubuntu 22.04 configuration for AMD MI300x

vanzod:
This PR introduces the following changes:

- Enables the use of the `AMD` flag to configure ROCm drivers and associated installations
- Automatically downloads and installs `amdgpu_install` utility
- Installs Cmake 3.30.5 required by RCCL
- Fixes the default assignment of user secondary groups `video` and `render`

### DIFF
--- a/ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
+++ b/ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -ex
 
+pushd /tmp
+wget https://github.com/Kitware/CMake/releases/download/v3.30.5/cmake-3.30.5-linux-x86_64.tar.gz
+tar xzf cmake-3.30.5-linux-x86_64.tar.gz
+pushd cmake-3.30.5-linux-x86_64
+pushd bin
+sudo mv -f ccmake cmake cpack ctest /usr/local/bin
+popd
+sudo cp -r share/cmake-3.30 /usr/local/share/
+popd
+rm -rf cmake-3.30.5-linux-x86_64*
+popd
+
 apt install libstdc++-12-dev
 apt remove -y rccl
 pushd ~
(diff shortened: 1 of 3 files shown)


hpourreza:
Looks like vv3.30.5 is no longer available. Could you please use https://github.com/Kitware/CMake/releases/download/v3.30.6/cmake-3.30.6-linux-x86_64.tar.gz

hpourreza:
Please also add `hash -r` after installing cmake to make sure rccl is using the correct version.

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -1,6 +1,18 @@
 #!/bin/bash
 set -ex
 
+pushd /tmp
+wget https://github.com/Kitware/CMake/releases/download/v3.30.5/cmake-3.30.5-linux-x86_64.tar.gz
```
hpourreza, line 5: Looks like vv3.30.5 is no longer available. Could you please use https://github.com/Kitware/CMake/releases/download/v3.30.6/cmake-3.30.6-linux-x86_64.tar.gz

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -1,6 +1,18 @@
...
+tar xzf cmake-3.30.5-linux-x86_64.tar.gz
+pushd cmake-3.30.5-linux-x86_64
+pushd bin
+sudo mv -f ccmake cmake cpack ctest /usr/local/bin
+popd
+sudo cp -r share/cmake-3.30 /usr/local/share/
+popd
+rm -rf cmake-3.30.5-linux-x86_64*
```
hpourreza, line 13: please add `hash -r` after removing the file to clear the stored path for cmake

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
LiquidPT: I strongly suggest using the common utility functions to retrieve the version from `versions.json`, verify your download, and write the component version used to the `component_versions.txt` so that it's stored in the image for reference. 

Most files have examples of doing this, but `https://github.com/Azure/azhpc-images/blob/master/common/install_azcopy.sh` is fairly simple. This should be done for all components that are being installed.

PR #392, opened 2025-01-29
Enable AMD GPU platform build 

hpourreza:
The scripts were updated to enable the build process on the AMD GPU platform


ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/bash
 set -ex
 
+#update CMAKE
+pushd /tmp
+CMAKE_VER="3.30.6"
+wget https://github.com/Kitware/CMake/releases/download/v${CMAKE_VER}/cmake-${CMAKE_VER}-linux-x86_64.tar.gz
```
moes1, line 7: Should this cmake install be in a different script?

hpourreza: should I add something like install_cmake.sh and call it from `install_utils.sh` or `install_prerequisites.sh` ?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/bash
 set -ex
 
+#update CMAKE
+pushd /tmp
+CMAKE_VER="3.30.6"
```
LiquidPT, line 6: As I called out in a PR a couple of weeks ago, I strongly suggest using the common utility functions to retrieve the version from versions.json, verify your download, and write the component version used to the component_versions.txt so that it's stored in the image for reference.

Most files have examples of doing this, but https://github.com/Azure/azhpc-images/blob/master/common/install_azcopy.sh is fairly simple. This should be done for all components that are being installed.

tests/run-tests.sh
```diff
@@ -67,10 +67,21 @@ function initiate_test_suite {
 }
 
 function set_test_matrix {
+    gpu_platform="NVIDIA"
+    if [[ "$#" -gt 0 ]]; then
+       INPUT=$1
```
LiquidPT, line 72: Should this have GPU in the name so that we know what it's supposed to be? GPU_INPUT or GPU_PARAM ?

tests/test-matrix_AMD.json
```diff
@@ -0,0 +1,22 @@
+{
+    "ubuntu22.04": {
+        "components": ["check_impi_2021", "check_rccl", "check_aocl", "check_aocc", "check_docker"],
```
LiquidPT, line 3: Ideally we'd have checks here to check that the ROCM drivers are installed correctly. I assume we don't because the install is happening on A100s?

tests/run-tests.sh
```diff
@@ -19,8 +19,9 @@ function test_component {
     case $component in
         check_impi_2021) verify_impi_2021_installation;;
         check_impi_2018) verify_impi_2018_installation;;
-        check_cuda) verify_cuda_installation;;
+        check_cuda) verify_gdrcopy_installation;verify_cuda_installation;;
```
LiquidPT, line 22: verify_gdrcopy_installation is called from verify_common_components. I you want it to be a switchable component, you'll need to remove it from there. Preferably it would have its own switch case here and be added to the NVIDIA configurations

tests/test-matrix_AMD.json
hpourreza: another reason is to know exactly what it means to verify rocm is correctly installed. I guess we can try checking the folder path, etc. which doesn't need to be on AMD platform

LiquidPT: I assume that there are some checks similar to what we do with CUDA that we could minimally do and add to later (install path check, an executable similar to `nvidia-smi` which will return version, modules showing in `lsmod`)

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
hpourreza: Trying to use the `download_and_verify.sh` file for cmake and rocm but this file does not take the path to store the downloaded file and calling something like this breaks it:
```
pushd /tmp
download_and_verify.sh <url> <sha256>
```
as it stands, I think I have to download the package to the source tree and remove it when it is done.

LiquidPT: usually we download into the working directory and then delete

tests/run-tests.sh
hpourreza: changed the variable to `GPU_PLAT`

tests/test-matrix_AMD.json
hpourreza: added some check that I think should work on non-amd VM as well

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
```diff
@@ -50,7 +69,7 @@ mkdir /opt/rocm-perftest
 pushd ~/rdma-perftest
 ./autogen.sh
 ./configure --enable-rocm --with-rocm=/opt/rocm --prefix=/opt/rocm-perftest/
-make -j 32
+make -j
```
LiquidPT, line 72: should this also not be `make -j$(nproc)`?

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
moes1: Does this cause problems if it's installed before this point in the installation process? If not then yeah it would probably be best to have it install in the install_prerequisites.sh script.

versions.json
```diff
@@ -234,6 +241,13 @@
             } 
         }
     },
+    "rocm": {
+        "ubuntu22.04": {
+            "version": "6.2.60204-1",
+            "url": "https://repo.radeon.com/amdgpu-install/6.2.4/ubuntu/jammy/amdgpu-install_6.2.60204-1_all.deb",
+            "sha256": "38ba7dbf74b2145e406d800772f7918b8cd53b3efb94f5f17cdea21497ef4dab"
+        }
+    },
```
darkwhite29, line 250: Thanks for adding the ROCm version. Could you please update the verstion to 6.2.4?

Also, the RCCL version built against ROCm 6.2.4 is 2.20.5 (https://github.com/ROCm/rccl/releases/tag/rocm-6.2.4). Could you please add one entry for RCCL as well?

darkwhite29 (changes requested):

versions.json
hpourreza: sure, I will add it. The naming is a bit confusing since the URL has reference to rocm 6.2.4 and downloaded file is called rccl-rocm-6.2.4.tar.gz. I will comment to clarify this is indeed rccl not rocm

ubuntu/ubuntu-22.x/ubuntu-22.04-hpc/install_rccl.sh
hpourreza: done

tests/run-tests.sh
hpourreza: I prefer not to touch the nvidia parts as much as possible.

tests/run-tests.sh
```diff
@@ -67,10 +68,21 @@ function initiate_test_suite {
...
+          gpu_platform="AMD"
+       elif [[ ${GPU_PLAT} != "NVIDIA" ]]; then
+          echo "${GPU_PLAT} is not a valid GPU platform"
+          exit 1
+
+       fi
+    fi
     export distro=$(. /etc/os-release;echo $ID$VERSION_ID)
-    test_matrix_file=$(jq -r . $HPC_ENV/test/test-matrix.json)
+    test_matrix_file=$(jq -r . $HPC_ENV/test/test-matrix_${gpu_platform}.json)
```
LiquidPT, line 83: please delete the now unused `test-matrix.json` file to avoid confusion

versions.json
```diff
@@ -234,6 +241,20 @@
             } 
         }
     },
+    "rocm": {
+        "ubuntu22.04": {
+            "version": "6.2.60204-1",
```
darkwhite29, line 246: Could you please update `version` to 6.2.4?

darkwhite29 (approved):

PR #394, opened 2025-02-04
updating url for azcopy download

egmsft:
The url to download azcopy in this repository no longer works. I have updated the URL so the install of azcopy is successful.


darkwhite29:
An earlier PR by a colleague has proposed the same update: https://github.com/Azure/azhpc-images/pull/392/files#diff-75849723f241364a8d80ef753d35bc39223022cbef74ab43d24c12f8835dc3d2

abhamidipati-msft:
Better to use package manager installation from PMC repo

egmsft:
I didn't realize this change was being addressed in another PR. I can close this one or test out an install via the PMC repo instead. Please let me know how you'd prefer to proceed.

PR #397, opened 2025-02-28
Fix render group membership

hpourreza:
The first username created during the VM deployment was not in the render group causing the rccl test not to run


darkwhite29 (approved):
LGTM.

PR #399, opened 2025-03-19
Update install_health_checks.sh

rajvi:


common/install_health_checks.sh
```diff
@@ -19,15 +19,16 @@ if [ "${GPU_PLAT}" = "NVIDIA" ]; then
 
    pushd azurehpc-health-checks
 
-   # Pull down docker container from MCR
-   ./dockerfile/pull-image-acr.sh cuda
+   # Pull CUDA Docker image from MCR
+   ./dockerfile/pull-image-mcr.sh cuda
    popd
 else
    git clone https://github.com/Azure/azurehpc-health-checks.git
```
ChrisRus, line 26: I notice that this git clone does not include the --branch as is present in the NVIDIA case. I think it's important that the image build always pull from a specific version.

common/install_health_checks.sh
```diff
@@ -19,15 +19,16 @@ if [ "${GPU_PLAT}" = "NVIDIA" ]; then
...
+   
    pushd azurehpc-health-checks
-   # Build docker image for AMD while waiting to be published on MCR
-   ./dockerfile/build_image.sh rocm
-
+   
+   # Pull ROCm Docker image from MCR
+   ./dockerfile/pull-image-mcr.sh cuda rocm
```
ChrisRus, line 31: The `pull-image-mcr.sh` script only parses a single in-parameter so I don't think this will pull the ROCm Docker as expected.

ChrisRus (changes requested):

common/install_health_checks.sh
rajvi: I agree. I decided not change this file much if the build images team has plans for something different within their project.

common/install_health_checks.sh
rajvi: Yes, I updated it.

common/install_health_checks.sh
ChrisRus: Consumers of the HPC Marketplace image will always expect that the NHC sources included in the image via git clone will exactly match the sources used to build the included Docker container image and that they're always included in HPC Marketplace image builds as a matched set.

From azhpc-image repo standpoint, I think it's important that NHC sources from GitHub are always pulled from a specific commit hash / branch name. And from NHC project standpoint that official Docker container releases published to MCR are based on sources in official release branches.

rajvi: Like I said before, I agree. I can add this on their behalf.

ChrisRus: It's on behalf of ensuring that the NHC drop included in the HPC Marketplace image has been tested, is consistent, and works as expected. I'm not sure how the inconsistency in the git clone was introduced here - I didn't look at the history. I was just noting that it doesn't seem to make sense.

The code change here is trivial. What's not entirely trivial is providing the image builder team with (a) the NHC GitHub release branch (b) a published Docker container that was built from the release sources, tested, and published to MCR so that the sources and Docker included in the image are consistent.

ChrisRus: Rajvi - I see you added the git clone --branch but think this will end up pulling 0.4.3 release of NHC which was released on 2024/10/24 and dated 7/22/24? Shouldn't there be another release that includes the current sources?

ChrisRus: @hpourreza - do you agree with my take on this? If we're going to include the NHC sources via git clone --branch then those sources should match or be very close to the sources used to build the included Docker container IMO.

rajvi: I created a new release and version.

PR #401, opened 2025-04-09
Downgrade rocm version for compatibility

hpourreza:
Dues to some compatibility issues, we need to downgrade the rocm version to 6.2.2


darkwhite29 (approved):
LGTM!

PR #404, opened 2025-04-29
Make sanity check continues on error

ljqqqqq:
Adds a flag on run-tests.sh and enables the sanity check continue to run when a single function fails, which can help understand the thorough VM status in local tests. The flag is optional.


darkwhite29 (approved):
LGTM!

PR #408, opened 2025-04-30
Fix RCCL

Shuyu-Zhao:
rccl_folder=$(tar -tzf $TARBALL | head -1 | sed 's:/$::')
The method needs to be used after the tar.gz file is downloaded, which will cause script fail.


darkwhite29 (approved):
Approve it on behalf of Matt.

PR #410, opened 2025-05-05
Use udev rule to grant AMD GPU access to all users

vanzod:
The proposed use of udev rules to grant all users access to GPUs follows the method described [in ROCm documentation](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/prerequisites.html#grant-gpu-access-to-all-users-on-the-system).
The previous method of relying on `/etc/adduser.conf` to assign all new users to `video,render` secondary groups fails if users are not created with `adduser`. This is for example the case of users created by CycleCloud.


darkwhite29 (approved):
LGTM!

PR #413, opened 2025-05-14
Do not upgrade impi and hpcx

ljqqqqq:
Do not upgrade the hpcx and impi version for both alma8.10 and ubuntu22.04 in this coming release. 
There are two reasons: 
1. For ubuntu22.04, in our already finished ADO validation pipelines, we observed serious performance downgrade with the latest version.
2. For alma8.10, in our already finished ADO validation pipelines, we observed the images with the latest hpcx and impi are stuck in mpi benchmarks.

So, we decided to keep the legacy version of hpcx and impi.


darkwhite29 (approved):
LGTM!

PR #426, opened 2025-07-26
Add Trivy vulnerability scanning

arsdragonfly:


common/install_azcopy.sh
Shuyu-Zhao, line 1: If we choose remove the azcopy installation script, we should also remove the azcopy section in versions.json

PR #436, opened 2025-09-19
add systemdcgroup=true on /etc/containerd/config.toml on azurelinux

ljqqqqq:
AzureLinux uses containerd 2.x, in which the config toml file is version 3. The default file generated by "containerd config default" does not have SystemdCgroup config 

image


Shuyu-Zhao (approved):

Shuyu-Zhao (commented):
Why change the config.toml after the runtime configure?

PR #458, opened 2025-12-08
Enable GB200 image build

ljqqqqq:
1. Change the structure of versions.json to "component -> distro -> architecture" to support arm64 image build.
2. Add tests to all distros: split "check_cuda" to "check_nvidia_drvier" and "check_cuda", add "check_nvlink" and "check_azure_persistent_rdma_naming"
3. Support ubuntu gb200 image build, including regular vm image and aks host image.

### DIFF
--- a/distros/ubuntu22.04/install.sh
+++ b/distros/ubuntu22.04/install.sh
@@ -28,7 +28,7 @@ source ../../utils/set_properties.sh
 $COMPONENT_DIR/install_cmake.sh
 
 # install Lustre client
-# $COMPONENT_DIR/install_lustre_client.sh
+$COMPONENT_DIR/install_lustre_client.sh
 
 # install DOCA OFED
 $COMPONENT_DIR/install_doca.sh
--- a/distros/ubuntu24.04/install.sh
+++ b/distros/ubuntu24.04/install.sh
@@ -27,7 +27,7 @@ if [ "$SKU" != "GB200" ]; then
 
     # install Lustre client
     # Note that lustre client is supported on GB200 but amlfs does not support latest 6.14 kernel so we temporarily skip it
-    # $COMPONENT_DIR/install_lustre_client.sh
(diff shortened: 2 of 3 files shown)


YutongSun (approved):

components/install_gdrcopy_aks.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
Shuyu-Zhao: Why we create a new file instead of adding a conditional check in the existing one?

distros/ubuntu24.04/install_aks.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
Shuyu-Zhao: Wouldn't it be better to separate ubuntu AKS specified scripts into a separate distro folder?

Shuyu-Zhao (dismissed):

utils/set_properties.sh
```diff
@@ -6,10 +6,15 @@ export COMPONENT_DIR=$TOP_DIR/components
 export TEST_DIR=$TOP_DIR/tests
 export UTILS_DIR=$TOP_DIR/utils
 export DISTRIBUTION=$(. /etc/os-release;echo $ID$VERSION_ID)
+export ARCHITECTURE=$(uname -m)
```
arsdragonfly: add extra variable that indicates distro-packaging-specific monikers (e.g. amd64 vs arm64 on Ubuntu)

ljqqqqq: I think x86_64 and arm64, which are exactly results of "uname -r", can effectively distinguish different branches in the installation of components. It mainly influence package url or repo, which may have a x86_64/arm64 suffix or prefix.

ljqqqqq: Please search "ARCHITECTURE" in the repo and check its usage

PR #473, opened 2026-01-16
add azure vm utils, plus some other fixes

### LINKED ISSUE #471: add azure-vm-utils in all images
CycleCloud release 8.8.x relies on the `azure-nvme-id` utility set in the /opt/cycle/jetpack/system/procedures/startup/80-azure-disk.rules rule file.
This utility is not installed on AL8 and AL9, but seems to be there in Ubuntu 24.04.

The utility can be installed from this repo https://github.com/Azure/azure-vm-utils
Make sure it will be there in all images, as it become a critical piece when discovering NVMe.

arsdragonfly:
fixes #471 
Also fixes AZL3 fabricmanager version alignment and adds ability to skip stale package check at validation pipeline

### DIFF
--- a/components/configure_nvidia_persistence.sh
+++ b/components/configure_nvidia_persistence.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+set -ex
+
+# Configure NVIDIA persistence daemon to keep the GPU driver loaded in memory
+# This eliminates cold start delays when launching GPU applications
+
+# Create systemd service file if it doesn't exist
+if [ ! -f /etc/systemd/system/nvidia-persistenced.service ]; then
+    cat <<EOF > /etc/systemd/system/nvidia-persistenced.service
+[Unit]
+Description=NVIDIA Persistence Daemon
+Wants=syslog.target
+ 
+[Service]
+Type=forking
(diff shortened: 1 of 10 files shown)


distros/almalinux8.10/install_utils.sh
```diff
@@ -101,6 +101,16 @@ yum localinstall ./dl.fedoraproject.org/pub/epel/8/Everything/x86_64/Packages/s/
 rm -rf ./dl.fedoraproject.org/
 rm -rf ./repo.almalinux.org/
 
+# Install azure-vm-utils from source (no upstream package available for AL8)
```
xpillons: On AL8 and AL9 I had to install these before.
```
dnf -y install dnf-plugins-core
dnf config-manager --set-enabled powertools
dnf -y install gcc make cmake pkgconf-pkg-config json-c json-c-devel
```
Also I ran cmake disabling the tests as it needs additional packages
```
cmake -DENABLE_TESTS=OFF .
```

arsdragonfly:
AZL3's upstream package is very [outdated](https://github.com/microsoft/azurelinux/issues/15661); AL9's upstream package seems to work fine.
We'll build from source for both AL8 and AZL3.

PR #483, opened 2026-02-05
feat: add support for Rocky Linux 8 and 9

gtorres-ciq:
This PR adds HPC image building support for Rocky Linux 8.10 and 9.6, extending Azure HPC platform compatibility to include RHEL-compatible Rocky Linux distributions.


arsdragonfly:
@gtorres-ciq sorry we just realized Rocky scripts are very botched... fixes will be in https://github.com/Azure/azhpc-images/pull/489

PR #510, opened 2026-04-07
 feat: Refactoring baremetal SKU support via env-var contract (GB200/GB300 + future SKUs)

MKGAURAB:
### Summary
 
 This PR adds scalable baremetal SKU support to azhpc-images, making the build and test pipeline work on non-Azure baremetal nodes (e.g. GB200/GB300) without hard-coding individual VM size strings throughout the
codebase.

Changes

Environment variable contract (utils/set_properties.sh)

 - Introduces NODE_TYPE env var (azure-vm default, baremetal for bare-metal nodes) to distinguish deployment contexts
 - Introduces SKU_FAMILY env var (e.g. gb-family) as a single canonical GPU-family identifier, derived from SKU or set directly by callers — replaces repeated per-SKU VMSIZE string comparisons across all downstream 
scripts
 - On baremetal, skips apt update/upgrade (no access to online mirrors); pins the correct kernel package per SKU family on Azure VMs

Test improvements (tests/test-definitions.sh, tests/run-tests.sh)

 - Replaces hardcoded VMSIZE comparisons with SKU_FAMILY checks throughout (NCCL, CDMM, NVLink, package upgrade, test matrix selection)
 - Splits verify_ib_device_status and verify_ib_modules_and_devices into baremetal vs Azure VM paths: baremetal expects IB LinkUp but no IPoIB; Azure VMs expect both
 - Improves NVLink Fabric health validation: previously only checked for literal N/A; now validates State=Completed, Status=Success, non-zero CliqueId, and valid ClusterUUID per GPU
 - Clarifies test output messages for IPoIB and IB link checks on baremetal
 - HPC_ENV and VMSIZE are now overridable via environment (no Azure IMDS dependency on baremetal)
 - Appends -baremetal to DISTRIBUTION for baremetal nodes so the test matrix can have separate entries

SKU customizations (components/setup_sku_customizations.sh)

 - Refactored to derive topoFamily from NCCL_TOPO_FAMILY (if pre-set) or IMDS lookup, then dispatch to the appropriate customization script — removes duplicated IMDS call and makes it bypassable on non-Azure nodes

DCGM (components/install_dcgm.sh, versions.json)

 - Removes the redundant dcgm4 version key and its baremetal aarch64 special-case logic; dcgm.ubuntu24.04.aarch64 already uses the same epoch versioning scheme

Other (versions.json)

 - Adds IMEX version entry for Ubuntu
  24.04 aarch64
 - Pins Moby version for Ubuntu
  24.04 aarch64 baremetal
 - Renames moby_engine key to moby for consistency
 
 ### Baremetal caller environment variables
 
 The following must be set before invoking the build/test scripts on a baremetal node:
 
 | Variable | Example value | Mandatory | Introduced in | Notes |
 |----------|---------------|-----------|---------------|-------|
 | `NODE_TYPE` | `baremetal` | Yes | `set_properties.sh` | Master discriminator - controls IB/IPoIB checks, apt upgrade skip, and IMDS skip |
 | `SKU` | `GB200` or `GB300` | Yes | `set_properties.sh` | Used to auto-derive `SKU_FAMILY`; must be set before sourcing the script |
 | `VMSIZE` | `standard_nd128isr_ndr_gb200_v6` | Yes | `run-tests.sh` | Pre-set to skip Azure IMDS query (unreachable on baremetal) |
 | `HPC_ENV` | `/home/hpcgb200` | Yes | `run-tests.sh` | Path to the HPC stack install; defaults to `/opt/azurehpc` if not set |
 | `NCCL_TOPO_FAMILY` | `gb-family` | Yes | `setup_sku_customizations.sh` | Pre-set at runtime to bypass IMDS lookup when configuring NCCL topology |
 | `SKU_FAMILY` | `gb-family` | No | `set_properties.sh` | Auto-derived from `SKU` (GB200/GB300 -> gb-family); override only if needed |


tests/run-tests.sh
```diff
@@ -132,29 +137,22 @@ function set_test_matrix {
...
-        if [[ ! -z "$component" ]]; then
-            eval "export $component" # Associates component name as variable and version as value
-        fi
-    done <<< "$component_versions"
+    export VERSION_OFED="25.10-1.7.1"
+    export VERSION_AZCOPY="10.31.1"
+    export VERSION_OMPI="5.0.8"
+    export VERSION_NVIDIA="580.105.08"
+    export VERSION_DOCKER="29.1.4-1"
+    export VERSION_NCCL="2.28.3-1"
```
hglkrijger: why revert to hardcoded values?

hglkrijger (approved):
generally looks fine. we'll need to make a few small changes in azure-dedicated-images after this goes in and we move forward, but I did not see anything concerning.

tests/run-tests.sh
MKGAURAB: Thanks for catching this. I thought we decided to keep the versions like that for baremetal; but actually we should not do it that way. Reverted that diff.

tests/run-tests.sh
```diff
@@ -79,7 +79,7 @@ function verify_common_components {
     verify_hpcx_installation;
     verify_ompi_installation;
     verify_pssh_installation;
-    if [[ "$VMSIZE" != "standard_nd128isr_ndr_gb200_v6" && "$VMSIZE" != "standard_nd128isr_gb300_v6" ]]; then
+    if [[ "${SKU_FAMILY:-}" != "gb-family" ]]; then
```
ColtonPaul, line 83: During installation we skip if baremetal, but during validation we skip if gb-family. I'd expect installation and validation to have identical conditionals

MKGAURAB: Actually this particular change is not directly related to baremetal. It is just consolidation and clean up of the logic based on gb-family condition which applies to Azure VM images as well. Amended the faulty logic in the installation phase.

tests/test-definitions.sh
```diff
@@ -46,13 +46,21 @@ function verify_ib_device_status {
     lspci | grep "Infiniband controller\|Network controller"
     check_exit_code "IB device is listed" "IB device not found"
 
-    # verify IB device is up
-    ibstatus | grep "LinkUp"
-    check_exit_code "IB device state: LinkUp" "IB link not up"
+    if [[ "${NODE_TYPE:-azure-vm}" == "baremetal" ]]; then
+        # Baremetal GB200/GB300: IB links are not brought up (NVLink is used instead)
+        ibstatus | grep "LinkUp"
```
ColtonPaul, line 51: Do we need `! ibstatus | grep "LinkUp"` ?

MKGAURAB: Looked at the original commit: https://github.com/Azure/azhpc-images/commit/e3d6b6cd7de11b60d9bf45c1f0985578d2229099. @Shuyu-Zhao not sure if we missed to invert the grep logic like Colton suggested. Was it expected? Otherwise I can introduce `!`

tests/test-definitions.sh
```diff
@@ -319,16 +335,28 @@ function verify_ib_modules_and_devices {
...
         check_exit_code "${module} module is inserted" "${module} module not inserted!"
     done
 
     # Check if ib devices are listed
-    ip addr | grep ib
-    check_exit_code "IPoIB is working" "IPoIB is not working!"
+    if [[ "${NODE_TYPE:-azure-vm}" == "baremetal" ]]; then
+        # On baremetal GB200/GB300, IPoIB is not used
+        ! (ip addr | grep ib)
+        check_exit_code "IPoIB is not working" "IPoIB is working!"
```
ColtonPaul: NIT: can we change the messages for clarity? Something like `check_exit_code "IPoIB not present as expected" "IPoIB unexpectedly present!"`

MKGAURAB: check_exit_code already gives the annotation for success and failure like below. Do we want to be more verbose ?

 - Success: prints [OK] : <message>
 - Failure: prints *** Error - <message>! + exit code, then calls exit_on_error

components/install_dcgm.sh
```diff
@@ -14,9 +14,17 @@ SKU_CUDA_VERSION=$(jq -r '.driver.version' <<< $cuda_metadata | cut -d'.' -f1)
 # Reference: https://developer.nvidia.com/dcgm#Downloads
 # the repo is already added during nvidia/ cuda installations
 
-# Get DCGM version from versions.json
-dcgm_metadata=$(get_component_config "dcgm")
-DCGM_VERSION=$(jq -r '.version' <<< $dcgm_metadata)
+# Get DCGM version from versions.json.
+# On baremetal aarch64, prefer the dcgm4-specific version key (different
+# epoch versioning scheme used in the baremetal package repository).
```
ColtonPaul: Could you double check if this is still necessary (including the addition of dcgm4 in versions.json)? In the nebius node branch, versions.json looked like
Image , but in latest master, we have a mix of version schemes all within `dcgm`.

utils/set_properties.sh
```diff
@@ -14,16 +14,42 @@ else
 fi
 export ARCHITECTURE=$(uname -m)
 
+# NODE_TYPE identifies the deployment context:
+#   'azure-vm'  (default) — Azure Virtual Machine image build/test
+#   'baremetal'           — Bare-metal node (e.g. Nebius), no Azure IMDS,
+#                           IB not brought up, IPoIB not used.
+# Baremetal callers must set NODE_TYPE=baremetal in their platform.conf
```
ColtonPaul: NIT: `platform.conf` is a thing from our internal user. We should just say "environment" in our public repo

components/install_dcgm.sh
MKGAURAB: I think we actually want the latest master one. I will address accordingly.

tests/test-definitions.sh
ColtonPaul: Wouldn't it be:
Success: `[OK] IPoIB is not working"` and
Failure: `... Error - IPoIB is working! ... `
?

Not suggesting more verbose, just the minor adjustments to the phrasing from my first comment. It's counterintuitive to see a [OK] message saying something isn't working and an error message saying that something *is* working.

ColtonPaul:
LGTM aside from comments. Thanks!

tests/test-definitions.sh
Shuyu-Zhao: No, we don't. We only disable IPoIB to avoid DHCP flood, not to disable IB Link, it should stay LinkUp.

tests/test-definitions.sh
MKGAURAB: Addressed in the latest iteration.

components/install_dcgm.sh
MKGAURAB: Addressed in the latest iteration.

utils/set_properties.sh
MKGAURAB: Addressed in the latest iteration.

Shuyu-Zhao (dismissed):
Seems fine

distros/ubuntu24.04/install_nvidiagpudriver_gb200.sh
```diff
@@ -69,7 +70,7 @@ write_component_version "NVIDIA" $nvidia_driver_version
 $COMPONENT_DIR/install_gdrcopy.sh
 
 # Install NVIDIA IMEX
-apt-get install nvidia-imex -y
+apt-get install nvidia-imex=${NVIDIA_IMEX_VERSION} -y --allow-downgrades
```
ljqqqqq: Curious why should we use "=${NVIDIA_IMEX_VERSION}", when we have already installed version pinning package?

MKGAURAB: I was retaining the commit from original author. However when confirmed what are the pkges get pinned down through the pinning pkg, l found nvidia-imex is also there. Hence this is not needed any more. Updated in the latest iteration.
image

Shuyu-Zhao (approved):

PR #516, opened 2026-04-17
Common packages upgrade

normanvuong:


versions.json
```diff
@@ -13,9 +13,9 @@
     },
     "cmake": {
```
arsdragonfly, line 14: This is probably leftover stuff from ages ago, CMake on all in-support distros should be modern enough to handle all compilation. Install CMake from respective distro repos instead of building from source even if source is newer

arsdragonfly (changes requested):

versions.json
```diff
@@ -714,7 +714,7 @@
     },
     "dcgm": {
         "common": {
-            "version": "3.3.9"
+            "version": "4.4.2"
```
ColtonPaul: I see u24 is up to 4.5.2, which is the latest. Does something require common to be a bit older? https://docs.nvidia.com/datacenter/dcgm/latest/release-notes/changelog.html#id2

normanvuong: The common is actually unused currently for dcgm as all other OS except RHEL has a specific version defined. I've updated it to 4.5.3 to reflect what's RHEL currently will install but will confirm to see if RHEL behavior is expected or not.

versions.json
normanvuong: Ah gotcha thank you for pointing that out. However, I still see all distro referencing the install_cmake script which reference this section. Refactor them to install from distro repos instead and deprecate the script might need a separate PR on its own. Is it okay if we only do the version bumping and follow up with your suggestion on a different PR?

PR #517, opened 2026-04-17
feat: update Azl 3.0 HPC component versioning

henryli001:
This PR updates the versioning of certain components in the Azure Linux 3.0 HPC image in terms of the following components:
- hpcx (v2.24.1 -> v2.25.1)
- hpcx_amd (v2.18 -> v2.23)
- mvapich (v2.3.7-2 -> v4.1)
- nccl (v2.28.3-1 -> v2.28.9-1)
- DCGM (v4.4.2 -> v4.5.2)
- nvshmem (v3.5.19 -> latest)

There are certain behavior changes when it comes to package installation in this change:
- Bump nvbandwidth in general from v0.8 to v0.9. The latest v0.9 contains an upstream [fix](https://github.com/NVIDIA/nvbandwidth/pull/51) that was made to resolve compilation failure on Azure Linux so that no extra patching is needed
- The current nvshmem RPMs used by Azure Linux HPC image are the RHEL packages built by NVIDIA. Thus, switch to directly consuming the latest RHEL RPMs from upstream NVIDIA RHEL cuda repo instead of the current one whose version is outdated.
- Modify cuda-toolkit package installation to consume the latest version from upstream NVIDIA Azl3.0 cuda repo instead of the current one whose version is pinned

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -150,19 +150,19 @@
         "azurelinux3.0": {
             "x86_64": {
                 "nvidia_v100":{
-                    "version": "2.24.1",
-                    "sha256": "e21eadd223541b887cf3cf4a4b21cd03c2bd867cfbf07ec00d64ac0411b63923",
-                    "url": "https://content.mellanox.com/hpc/hpc-x/v2.24.1_cuda12/hpcx-v2.24.1-gcc-doca_ofed-redhat9-cuda12-x86_64.tbz"
+                    "version": "2.25.1",
+                    "sha256": "006b976fd64e79eb26177fe45714252e78d0e09d27b8cdbf1baf7be6fee8fa32",
+                    "url": "https://content.mellanox.com/hpc/hpc-x/v2.25.1_cuda12/hpcx-v2.25.1-gcc-doca_ofed-redhat9-cuda12-x86_64.tbz"
             
                 },
-                "version": "2.24.1",
-                "sha256": "d2bf870830c88d096a209fc33bb7af7643e755b660e43fe60a3e6139af1d3016",
-                "url": "https://content.mellanox.com/hpc/hpc-x/v2.24.1_cuda13/hpcx-v2.24.1-gcc-doca_ofed-redhat9-cuda13-x86_64.tbz"
+                "version": "2.25.1",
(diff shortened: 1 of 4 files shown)


versions.json
```diff
@@ -218,9 +218,9 @@
         },
         "azurelinux3.0": {
             "x86_64": {
-                "version": "2.18",
-                "sha256": "45276ff7bd676cc668d1cc6a1fe926d5e157646aaf06201415e0aadb048be16d",
-                "url": "https://content.mellanox.com/hpc/hpc-x/v2.18/hpcx-v2.18-gcc-mlnx_ofed-redhat8-cuda12-x86_64.tbz"
+                "version": "2.23",
```
arsdragonfly, line 221: this was held back due to perf regression, do not bump yet

versions.json
```diff
@@ -253,9 +253,9 @@
         },
         "azurelinux3.0": {
             "x86_64": {
-                "version": "2.3.7-2",
-                "sha256": "03fc160c81a1d522b1e7c6723841e27fcab04d527484e4242d3e66d7c5024205",
-                "url": "http://mvapich.cse.ohio-state.edu/download/mvapich/mv2/mvapich2-2.3.7-2.tar.gz"
+                "version": "4.1",
+                "sha256": "25a53d3725b669e2c648158fb7c9fc5b1388953f3a2f949748586c447d0e43ee",
+                "url": "https://mvapich.cse.ohio-state.edu/download/mvapich/mv2/mvapich-4.1.tar.gz"
```
arsdragonfly, line 258: mvapich 4 has perf regressions, do not bump if 2.3.7 continues to build

versions.json
henryli001: Are we planning to roll back the version on other distros as well, say Ubuntu 24.04?

versions.json
henryli001: Same question as the one for the other comment

versions.json
arsdragonfly: We observed azl3 specific regression, hence the holdback (and the decision to bifurcate hpcx_amd from hpcx in the first place)

versions.json
arsdragonfly: We observed azl3-specific regression

PR #520, opened 2026-04-28
Add support for NCv6 SKU 

ColtonPaul:
- Install GRID driver instead of standard Nvidia driver
- Replace UCX with libfabric as MPI transport for non-IB SKUs for compatibility with MANA
- Use HPC-X inbox build
- Blacklist mana_ib until it's fully functional
- Skip mlnx-specific libraries for non-ib skus (currently just NCv6)


tests/test-definitions.sh
```diff
@@ -435,6 +468,9 @@ function verify_nvloom_setup {
 }
 
 function verify_nvlink_setup {
+    # Skip NVLink checks on SKUs without NVLink
+    if ! has_infiniband; then return; fi
```
YutongSun: has_nvlink instead of has_infiniband?

components/install_nvidiagriddriver.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Install CUDA toolkit
+cuda_metadata=$(get_component_config "cuda")
+CUDA_DRIVER_VERSION=$(jq -r '.driver.version' <<< $cuda_metadata)
+CUDA_DRIVER_DISTRIBUTION=$(jq -r '.driver.distribution' <<< $cuda_metadata)
+# Add NVIDIA CUDA APT repo (provides toolkit packages)
+wget https://developer.download.nvidia.com/compute/cuda/repos/${CUDA_DRIVER_DISTRIBUTION}/x86_64/cuda-keyring_1.1-1_all.deb
+dpkg -i ./cuda-keyring_1.1-1_all.deb
+apt-get update
+apt install -y cuda-toolkit-${CUDA_DRIVER_VERSION//./-}
```
YutongSun, line 41: distro check for Ubuntu, may support other distros

components/install_mpis.sh
```diff
@@ -29,16 +32,21 @@ download_and_verify ${HPCX_DOWNLOAD_URL} ${HPCX_SHA256}
 tar -xvf ${TARBALL}
 
 sed -i "s/\/build-result\//\/opt\//" ${HPCX_FOLDER}/hcoll/lib/pkgconfig/hcoll.pc
+sed -i "s/\/build-result\//\/opt\//" ${HPCX_FOLDER}/ucx/lib/pkgconfig/*.pc
 mv ${HPCX_FOLDER} ${INSTALL_PREFIX}
 HPCX_PATH=${INSTALL_PREFIX}/${HPCX_FOLDER}
 HCOLL_PATH=${HPCX_PATH}/hcoll
 UCX_PATH=${HPCX_PATH}/ucx
+LIBFABRIC_PATH=/opt/libfabric
```
YutongSun, line 40: it is set for all skus

tests/test-definitions.sh
ColtonPaul: Agreed, added new helper function

components/install_nvidiagriddriver.sh
ColtonPaul: This script is currently only called by the Ubuntu 24.04, so no need for a distro check here. But good point that there's nothing that fast-fails for unsupported SKU+OS combinations. I have ideas for this, but I don't think it'll fold in nicely to this PR, so I'd like to keep it separate -- it may require restructuring the pipelines + the install scripts to add a layer of indirection for a central entry point that can do the checks.

components/install_mpis.sh
ColtonPaul: This is just variable assignment, so it's harmless. I don't think it's worth the extra lines to wrap it in a check.

tests/run-tests.sh
```diff
@@ -68,12 +68,17 @@ function test_component {
 # Verify common component installations accross all distros
 function verify_common_components {
     # Skip package updates check in validation mode (only run at build time)
-    if [[ -z "${validation_mode:-}" ]]; then
-        verify_package_updates;
+    verify_package_updates;
+    if has_infiniband; then
```
MKGAURAB: Why splitting in two different if conditions and not all three verify under one ìf has_infiniband` condition?

tests/run-tests.sh
```diff
@@ -68,12 +68,17 @@ function test_component {
 # Verify common component installations accross all distros
 function verify_common_components {
     # Skip package updates check in validation mode (only run at build time)
```
MKGAURAB, line 70: Maybe we want to remove this comment as well, now that we have remove the condition for `validation_mode`?

components/install_nvidiagpudriver.sh
```diff
@@ -114,26 +114,16 @@ if [[ "$DISTRIBUTION" != *-aks ]]; then
...
-    CUDA_SAMPLES_DOWNLOAD_URL=https://github.com/NVIDIA/cuda-samples/archive/refs/tags/${TARBALL}
-    download_and_verify ${CUDA_SAMPLES_DOWNLOAD_URL} ${CUDA_SAMPLES_SHA256}
-    tar -xvf ${TARBALL}
-    pushd ./cuda-samples-${CUDA_SAMPLES_VERSION}
-    mkdir build && cd build
-    cmake -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc ..
-    make -j $(nproc)
-    mv -vT ./Samples /usr/local/cuda-${CUDA_DRIVER_VERSION}/samples # Use the same version as the CUDA toolkit as thats where samples is being moved to
-    popd
+    $COMPONENT_DIR/install_cuda_samples.sh
```
MKGAURAB, line 126: thank you for refactoring it out.

tests/run-tests.sh
ColtonPaul: Thanks for calling out the inconsistency. I had originally made a fix here that I then undid in favor of upstream changes. I intended to revert this hunk as well.

tests/run-tests.sh
ColtonPaul: Oops, thanks!

MKGAURAB (approved):

PR #524, opened 2026-05-04
GB200/300 April release

ljqqqqq:
1. Version upgrade: gdrcopy, nvloom, doca
2. Enable kernel selection on GB200
3. Purge kernel 6.17 on GB200
4. Refine image replication region on GB200

### DIFF
--- a/packer/scripts/prerequisites.sh
+++ b/packer/scripts/prerequisites.sh
@@ -173,12 +173,12 @@ install_ubuntu_lts_kernel() {
             # Purge non-LTS kernels
             apt-get purge -y \
                 linux-azure linux-image-azure \
-                "linux-image-6.11*" "linux-image-6.14*" \
-                "linux-azure-6.11*" "linux-azure-6.14*" \
-                "linux-cloud-tools-6.11*" "linux-cloud-tools-6.14*" \
-                "linux-headers-6.11*" "linux-headers-6.14*" \
-                "linux-modules-6.11*" "linux-modules-6.14*" \
-                "linux-tools-6.11*" "linux-tools-6.14*"
+                "linux-image-6.11*" "linux-image-6.14*" "linux-image-6.17*" \
+                "linux-azure-6.11*" "linux-azure-6.14*" "linux-azure-6.17*" \
+                "linux-cloud-tools-6.11*" "linux-cloud-tools-6.14*" "linux-cloud-tools-6.17*" \
+                "linux-headers-6.11*" "linux-headers-6.14*" "linux-headers-6.17*" \
+                "linux-modules-6.11*" "linux-modules-6.14*" "linux-modules-6.17*" \
+                "linux-tools-6.11*" "linux-tools-6.14*" "linux-tools-6.17*"
             
             apt autoremove -y
             apt upgrade -y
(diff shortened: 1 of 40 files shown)


arsdragonfly (approved):

versions.json
```diff
@@ -725,7 +725,7 @@
                         "commit": "0f7366e73b019e7facf907381f6b0b2f5a1576e4",
                         "distribution": "Ubuntu24_04"
                     }
-                },
+                },
```
ColtonPaul, line 728: Nit: accidental whitespace addition

packer/scripts/prerequisites.sh
```diff
@@ -100,15 +100,23 @@ install_ubuntu_gb200_kernel() {
     export NEEDRESTART_MODE=a
     
     apt-get update
-    apt-get install -y linux-azure-nvidia
-    apt-mark hold linux-azure-nvidia
+    if [[ "${KERNEL_VERSION}" == "6.14" ]]; then
+        sudo apt-get install linux-azure-nvidia -y
```
ColtonPaul, line 104: Nit: We can remove `sudo` for these new lines

packer/scripts/prerequisites.sh
```diff
@@ -100,15 +100,23 @@ install_ubuntu_gb200_kernel() {
...
+        sudo apt-mark hold linux-azure-nvidia-6.8                
+    fi
     
     # Purge non-nvidia kernels
     apt-get purge -y linux-azure linux-image-azure
 
     # Remove non-nvidia kernel packages
     local packages_to_remove
-    packages_to_remove=$(dpkg -l | awk '/linux-(azure|image|cloud-tools|headers|modules|tools)-6\.14/ && $2 !~ /nvidia/ {print $2}' || true)
+    packages_to_remove=$(dpkg -l | awk '/linux-(azure|image|cloud-tools|headers|modules|tools)-6\.(14|17)/ && $2 !~ /nvidia/ {print $2}' || true)
```
ColtonPaul, line 119: does this need to cover 6.8 as well?

utils/set_properties.sh
```diff
@@ -43,7 +43,7 @@ if [[ $DISTRIBUTION == *"ubuntu"* ]]; then
         # Azure VM: pin the kernel package to prevent unintended kernel upgrades,
         # then upgrade all other pre-installed components.
         if [[ "${SKU_FAMILY}" == "gb-family" ]]; then
-            apt-mark hold linux-nvidia-64k-hwe-24.04
+            apt-mark hold linux-azure-nvidia
```
ColtonPaul, line 46: Do we need this line since we already hold in prerequisites.sh? And if so, does it need to branch on the version name?

versions.json
ljqqqqq: Fixed

packer/scripts/prerequisites.sh
ljqqqqq: Fixed

packer/scripts/prerequisites.sh
ljqqqqq: The kernel purge is intended to remove the original kernel packages brought by the base image. The system is already running 6.17 (for example, 6.17.0-1013-azure), so 6.8 is no longer needed here. Actually, 6.14 is also no longer required, but we are keeping it for safety.

PR #533, opened 2026-05-13
Support Bare Metal H100 Images

ColtonPaul:
- Add A100|H100|H200 -> A100plus case to set_properties.sh
- Add A100plus entry to test-matrix_NVIDIA.json
- Widen moby pinning to all baremetal, add x86_64 moby version


MKGAURAB (approved):
LGTM

PR #539, opened 2026-05-21
Install g2 root CA to access download.microsoft.com

normanvuong:
Microsoft has rotated public-facing endpoints (e.g. `download.microsoft.com`) to a chain issued by a new intermediate, `Microsoft TLS G2 RSA CA OCSP NN`, whose issuer is `Microsoft TLS RSA Root G2`. That root exists in two forms with the same Subject and the same public key but different issuers: a self-signed root (valid until 2040) and a cross-signed certificate that is itself issued by DigiCert Global Root G2 and therefore behaves as an intermediate. 

The acute problem is a server-side chain-configuration inconsistency at Microsoft, not a missing trust anchor in the distro: some Microsoft front-ends serve the full chain (leaf → intermediate → the cross-signed root), which validates out of the box because DigiCert Global Root G2 is already in every supported distro's trust store; other front-ends omit the cross-signed bridge and serve only the leaf and intermediate, so the client must resolve the issuer of that intermediate locally, and unless the self-signed `Microsoft TLS RSA Root G2` is present in the trust store, path-building dead-ends and openssl/curl/etc. report `unable to get local issuer certificate` (verify error 20). 

Installing the self-signed root as a local trust anchor closes that gap: it terminates the chain locally when the server omits the cross-signed bridge, and is a no-op when the server does return the full chain (the existing DigiCert path is preferred). The remediation fetches the root over plain HTTP from Microsoft's PKI repository, pins its SHA-256 fingerprint, and installs it using the distro-appropriate layout — ca-certificates + `update-ca-certificates` on Ubuntu, and `/etc/pki/ca-trust/source/anchors/` + `update-ca-trust` on the RHEL family and Azure Linux. Once Microsoft fixes the partial-chain endpoints (or otherwise stops relying on the new G2 CA), this anchor becomes redundant but harmless.


components/install_microsoft_tls_root_g2.sh
```diff
@@ -0,0 +1,113 @@
...
+    ANCHOR_DIR="/etc/pki/ca-trust/source/anchors"
+    ANCHOR_FILE="${ANCHOR_DIR}/${ANCHOR_NAME}.pem"
+    TRUST_BUNDLE="/etc/pki/tls/certs/ca-bundle.crt"
+    update_trust() { update-ca-trust extract; }
+else
+    echo "Unsupported DISTRIBUTION='${DISTRIBUTION:-unset}'; refusing to install Microsoft TLS RSA Root G2 anchor."
+    exit 1
+fi
+
+# Idempotency: if a previous image bake already installed the anchor, exit.
```
ColtonPaul: nit: these idempotency lines are in case the base image already contains the cert, right? "Previous image bake" is a bit unclear to me

normanvuong: I have edit the comment to make it more clear :)

distros/ubuntu24.04/install_utils.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -ex
 
+# Install the "Microsoft TLS RSA Root G2" trust anchor before any HTTPS
```
ColtonPaul, line 4: Nit: Since the comments are duplicated across 7 files, "Install the "Microsoft TLS RSA Root G2" trust anchor before any HTTPS calls to Microsoft endpoints." might be sufficient since you can look inside the script for details

components/install_microsoft_tls_root_g2.sh
```diff
@@ -0,0 +1,113 @@
...
+#     comment headers, where `grep` would match -- but relying on that
+#     would silently regress on Ubuntu.
+#
+# Walk the bundle through openssl so the check is identical on every
+# supported distro: crl2pkcs7 wraps the PEM blocks into a single PKCS#7
+# envelope, then `pkcs7 -print_certs -noout` renders one stable
+# "subject=... issuer=..." line per certificate that we can grep.
+openssl crl2pkcs7 -nocrl -certfile "${TRUST_BUNDLE}" \
+    | openssl pkcs7 -print_certs -noout \
+    | grep -q "Microsoft TLS RSA Root G2"
```
ColtonPaul: Clever!

Could we wrap this in an `if` with a log message --> exit 1 ? As is, this will correctly fail the script, but it'll do so pretty subtlely

ColtonPaul:
LGTM with just a couple suggestions. Thanks!

components/install_microsoft_tls_root_g2.sh
normanvuong: ah good catch I have added it :)

distros/ubuntu24.04/install_utils.sh
normanvuong: Good point! I have edit the comments per suggestion!

ColtonPaul (approved):

PR #541, opened 2026-05-28
add RHEL Lustre build-from-source

normanvuong:
Enable lustre build from source for RHEL family (except AZL3)


components/install_nvidia_fabric_manager.sh
```diff
@@ -55,7 +55,10 @@ else
     
     yum install -y ./${FILENAME}
 
-    # Prevent package from being updated after installation
-    sed -i "$ s/$/ ${PACKAGE_NAME}/" /etc/dnf/dnf.conf
+    # Prevent package from being updated after installation. Critical: FM
+    # version must match the NVIDIA driver version exactly, so a stray
+    # 'yum update -y' (e.g. from install_nvidia_container_toolkit.sh) must
+    # not pull in the latest fabricmanager from the cuda-rhel9 repo.
```
ColtonPaul: I don't follow why this comment is added here specifically?

distros/rocky9.x/install_utils.sh
```diff
@@ -141,12 +141,19 @@ rm -f environment-modules-5.3.0-1.el9.x86_64.rpm
...
-sed -i "$ s/$/ shim*/" /etc/dnf/dnf.conf
-sed -i "$ s/$/ grub2*/" /etc/dnf/dnf.conf
+## Disable kernel updates (skipped when building Lustre from source or when
+## refreshing an image in-place so that DKMS-style rebuilds can keep up with
+## kernel upgrades, matching the Ubuntu prerequisites.sh behavior). The
+## shim*/grub2* sed lines extend the just-added exclude= directive, so they
+## must stay inside the same conditional -- running them without an exclude
+## line would corrupt the previous last line of dnf.conf (e.g.
+## skip_if_unavailable=False).
+if [[ "${LUSTRE_BUILD_FROM_SOURCE,,}" != "true" && "${REFRESH_MODE,,}" != "true" ]]; then
```
ColtonPaul, line 151: Do we only plan on supporting REFRESH_MODE=true on images that were created with LUSTRE_BUILD_FROM_SOURCE=true? In other words, does anything need to possibly remove the kernel exclusion from dnf.conf?

distros/rocky8.10/install_utils.sh
```diff
@@ -89,12 +95,19 @@ rm -f environment-modules-4.5.2-4.el8.x86_64.rpm
 ## Install kernel-abi-stablelists (needed by DOCA) before locking kernel packages
 dnf install -y kernel-abi-stablelists
 
-## Disable kernel updates (but not kernel-rpm-macros and other tools)
-echo "exclude=kernel kernel-core kernel-modules kernel-devel kernel-headers kernel-modules-extra" | tee -a /etc/dnf/dnf.conf
-
-# Disable dependencies on kernel core
-sed -i "$ s/$/ shim*/" /etc/dnf/dnf.conf
-sed -i "$ s/$/ grub2*/" /etc/dnf/dnf.conf
+## Disable kernel updates (skipped when building Lustre from source or when
```
ColtonPaul, line 98: Would it be possible to share the new duplicate code between the two rocky install_util.sh changes?

components/install_lustre_client.sh
```diff
@@ -78,6 +85,79 @@ elif [[ $DISTRIBUTION == *"ubuntu"* ]]; then
         echo "##[warning]No Lustre client packages available for kernel minor version ${KERNEL_MINOR}. Skipping Lustre installation."
         exit 0
     fi
+elif [[ $LUSTRE_BUILD_FROM_SOURCE == "true" ]]; then
+    # RHEL-family build-from-source: mirrors the Ubuntu flow above so AMLFS
```
ColtonPaul: I'm struggling to follow the AI-generated comments on this PR. I think they're a bit too verbose while at the same time kind of terse and jargony. I think even asking the AI itself to rewrite its comments in this PR could go a long way.

"Same arsdragonfly/dkms-* branch -- its patches only touch Debian/* and a clang CFLAGS toggle, so the RPM path is unchanged."
- I'm not actually sure why this is called out (and it's subject to change if Zheyu changes his branch), but it took a while to figure out that it's just saying Ubuntu and RHEL can use the same lustre branch.

"Base toolchain (gcc/autotools/rpm-build/kernel-*-devel/dkms) comes from distros/<rhel>/install_utils.sh; only the lustre-specific libs are added below."
- = Lustre's standard build tools are already installed, so just install lustre-specific dependencies

"openmpi-devel is intentionally skipped: it's only a BuildRequires of the lustre-tests subpackage (disabled via --disable-tests below) and would conflict with DOCA's openmpi 4.1.9a1 (strict-pinned by clusterkit on EL9 / opensm-libs on EL8) -- appstream openmpi-devel needs the libmpi.so.40()(64bit)(openmpi-x86_64) namespaced provide that DOCA's openmpi doesn't carry, aborting dnf install before lustre builds"
- = We intentionally skip openmpi-devel because it's only used by lustre-tests, which we disable anyway, and installing it breaks DOCA's strict dependency on its own OpenMPI package

tests/run-tests.sh
```diff
@@ -110,6 +110,45 @@ function initiate_test_suite {
...
+# Note: This script is invoked by the Packer "shell" provisioner as the
+# non-root build user (e.g. hpcuser), so any state-changing systemctl call
+# must go through sudo or polkit will reject it with "Interactive
+# authentication required".
+function ensure_nvidia_fabricmanager_active {
+    # Match the same SKU set used by verify_nvidia_fabricmanager_service:
+    # NDv4 A100 (NVSwitch) and NDv5 H100/H200 (NVSwitch).
+    local valid_sizes="standard_nd96.*v4|standard_nd96is.*_h[12]00_v5"
+    if ! [[ "${VMSIZE}" =~ ^($valid_sizes)$ ]]; then
+        return 0
```
ColtonPaul, line 128: Could we break this into a little util instead of duplicating code that we might need to edit in a couple places if it changes? There could be a `sku_has_nvswitch` similar to things like `sku_has_infiniband`

tests/run-tests.sh
```diff
@@ -240,6 +279,11 @@ set_module_files_path
 set_component_versions
 # Set current SKU and distro
 set_vm_properties $aks_host_flag
+# Ensure nvidia-fabricmanager is up on NVSwitch SKUs so CUDA can initialize
+# during validation. No-op for AMD platform and non-NVSwitch SKUs.
```
ColtonPaul: Nit: unnecessary comment

I get that it follows the pattern in the file, but I think we should do away with the pattern. Like, `# Set test matrix` is a comment for `set_test_matrix`

tests/test-definitions.sh
```diff
@@ -458,8 +458,11 @@ function verify_dcgm_installation {
 }
 
 function verify_sku_customization_service {
-    # Check if the SKU customization service is active
-    local valid_sizes="standard_nc.*ads_a100_v4|standard_nd96.*v4|standard_nd40rs_v2|standard_hb176.*v4|standard_nd96is*_h100_v5|standard_nc.*_rtxpro6000bse_v6"
+    # Check if the SKU customization service is active.
+    # Note: bash =~ is ERE, so 'nd96is.*_h[12]00_v5' is the regex equivalent of
```
ColtonPaul: NIT: I think we can simplify the comment to just
```
# Check if the SKU customization service is active.
# Note: bash =~ is ERE, so use regex instead of glob patterns for matching
```

tests/test-definitions.sh
```diff
@@ -468,8 +471,10 @@ function verify_sku_customization_service {
 }
 
 function verify_nvidia_fabricmanager_service {
-    # Check if the NVIDIA Fabricmanager service is active
-    local valid_sizes="standard_nd96.*v4|standard_nd96is*_h100_v5"
+    # Check if the NVIDIA Fabricmanager service is active.
+    # Covers NDv4 A100 (NVSwitch) and NDv5 H100/H200 (NVSwitch). See note in
+    # verify_sku_customization_service for the regex form.
```
ColtonPaul: Nit: I don't think we need a reference to the past comment

components/install_doca.sh
```diff
@@ -106,6 +106,26 @@ else
     dnf -y install doca-ofed
     # Restore exclusion
     mv /etc/dnf/dnf.conf.bak /etc/dnf/dnf.conf
+
```
arsdragonfly, line 99: Is there a way to fix the "stale packages" that we are seeing by deny-listing `mft` from CUDA repo (i.e. only allowing `mft` from DOCA repo)? We'd like to avoid unnecessarily pinning packages

ColtonPaul (commented):
Looks good, just have a few requests about code comments / reused functions

utils/utilities.sh
```diff
@@ -134,3 +134,46 @@ function sku_has_infiniband {
...
+# 'yum update -y' upgrade pinned packages -- most critically
+# nvidia-fabricmanager, which must match the NVIDIA driver version exactly.
+#
+# @Args     : One or more package names/globs (e.g. "ucx*" "openmpi").
+# @Returns  : 0 on success, 0 (no-op) if /etc/dnf/dnf.conf is absent.
+############################################################################
+function dnf_pin_packages {
+    [[ -f /etc/dnf/dnf.conf ]] || return 0
+    local pkgs=("$@")
+    [[ ${#pkgs[@]} -gt 0 ]] || return 0
```
ColtonPaul: Would we ever intentionally call this with no arguments, or would that signal an accidental error, like referring to an unset var?

components/install_lustre_client.sh
```diff
@@ -78,6 +85,79 @@ elif [[ $DISTRIBUTION == *"ubuntu"* ]]; then
...
+    # kmod ships as a DKMS RPM that auto-rebuilds against future host kernels
+    # via `dnf upgrade kernel`. Same arsdragonfly/dkms-* branch -- its patches
+    # only touch debian/* and a clang CFLAGS toggle, so the RPM path is
+    # unchanged. Base toolchain (gcc/autotools/rpm-build/kernel-*-devel/dkms)
+    # comes from distros/<rhel>/install_utils.sh; only the lustre-specific
+    # libs are added below.
+    #
```
arsdragonfly, line 93: Can we also do a marker package trick (like Ubuntu) or patch out the `Requires/BuildRequires` and then load HPCX, for `lustre-tests` to work?

arsdragonfly (changes requested):

components/install_nvidia_fabric_manager.sh
normanvuong: Good catch. I've removed it in: https://github.com/Azure/azhpc-images/pull/546

utils/utilities.sh
normanvuong: good catch. I have fixed this in https://github.com/Azure/azhpc-images/pull/546 to raise error if we called this with no arguments

tests/test-definitions.sh
normanvuong: fixed in: https://github.com/Azure/azhpc-images/pull/546

tests/test-definitions.sh
normanvuong: fixed in: https://github.com/Azure/azhpc-images/pull/546

tests/run-tests.sh
normanvuong: I have added the sku_has_nvswitch function in: https://github.com/Azure/azhpc-images/pull/546

tests/run-tests.sh
normanvuong: fixed in: https://github.com/Azure/azhpc-images/pull/546

components/install_doca.sh
normanvuong: I have swapped out this pin solution with alternative: https://github.com/Azure/azhpc-images/pull/546.

distros/rocky9.x/install_utils.sh
normanvuong: --In other words, does anything need to possibly remove the kernel exclusion from dnf.conf?-- Nope, not currently, only image that already has LUSTRE_BUILD_FROM_SOURCE=true will be able to have the kernel hold removed.

components/install_lustre_client.sh
normanvuong: yes I have switch to marker package solution: https://dev.azure.com/hpc-platform-team/hpc-image-val/_build/results?buildId=32707&view=results

components/install_lustre_client.sh
```diff
@@ -78,6 +85,277 @@ elif [[ $DISTRIBUTION == *"ubuntu"* ]]; then
...
+        flex
+
+    # Build and install an empty marker RPM that does two things:
+    #   (a) Provides openmpi-devel, so rpmbuild's BuildRequires check
+    #       passes without dnf trying to install the appstream package.
+    #   (b) Owns /usr/lib64/lustre-tests-mpi, the directory where the
+    #       MPI-linked lustre test binaries will be installed (set up
+    #       by the rpm macros below). Without an owning package the
+    #       directory would be unowned in the final image -- cosmetic,
+    #       but rpm-build flags it as a warning.
```
arsdragonfly, line 121: Add a `TODO` marking this as a temporary dirty fix, since a proper fix would eventually need to land in Lustre's build scripts

components/install_lustre_client.sh
```diff
@@ -78,6 +85,277 @@ elif [[ $DISTRIBUTION == *"ubuntu"* ]]; then
...
+    # step above does the equivalent in one call).
+    #
+    # Why the split (dnf install for userland+dkms, rpm -ivh --nodeps
+    # for the test kmod):
+    #
+    # There are TWO layers of rpm-level `ksym(symbol) = <hash>`
+    # auto-generated Requires that bite this lustre+DOCA OFED build,
+    # and the workaround for each layer is different.
+    #
+    # Layer 1 -- static kmod-lustre-client requires OFED-provided
```
arsdragonfly: Layer 1's description can be simplified to "DKMS over prebuilt kmod".
Layer 2's description can be "`kmod-lustre-client-tests`, which is a collection of kmods as test fixtures, lacks DKMS", which hopefully can be solved at the Lustre source one day...

arsdragonfly (dismissed):

components/install_lustre_client.sh
normanvuong: thanks for the suggestion. I have edit the comment to make it less redundant and jargony.

arsdragonfly (approved):

PR #543, opened 2026-05-31
Always write kernel version to component_versions

ColtonPaul:
- Previously, the kernel version was only written to `component_versions.txt` during Azure Health Checks installation which is sometimes skipped. Write the version in a dedicated step instead.

### DIFF
--- a/components/install_health_checks.sh
+++ b/components/install_health_checks.sh
@@ -36,12 +36,3 @@ popd
 popd
 
 write_component_version "AZ_HEALTH_CHECKS" ${AZHC_VERSION}
-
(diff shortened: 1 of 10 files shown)


ColtonPaul:
Tested by creating a new NCv6 image off of this commit and verifying that, unlike the last released preview version, it does have KERNEL in component_versions.txt
```
user@vm-05-29-2580-15c6:~$ cat /opt/azurehpc/component_versions.txt | grep -i kernel
  "KERNEL": "6.8.0-1056-azure",
```

MKGAURAB (approved):

PR #546, opened 2026-06-01
Update rhel family 9.7->9.8, fix in-place refresh kernel hold

normanvuong:
### Summary

Bumps the EL9 family (AlmaLinux & Rocky) from 9.7 to 9.8, reorganizes the per-distro install scripts so 9.x minors share a single folder, and fixes a kernel-hold bug that prevented in-place image refreshes from picking up new kernels. Also lands a set of DNF pinning / repo-exclude hardening fixes that surfaced while bringing 9.8 up.

### Motivation

- AlmaLinux 9.7 and Rocky Linux 9.7 are end-of-life on the upstream mirrors; 9.8 is now the supported EL9.x minor.
- The 9.7 → 9.8 baseos brought a brand-new RDMA-core stack (libhns / IBVERBS_1.15) that conflicts with DOCA's libibverbs, which broke `install_pmix.sh`'s `yum update -y` and the DOCA OFED install path.
- The legacy `sed -i "$ s/$/ PKG/" /etc/dnf/dnf.conf` pinning pattern silently corrupted the last line of `dnf.conf` (e.g. `skip_if_unavailable=False`) on configurations where no `exclude=` line had been seeded, letting pinned packages — most critically `nvidia-fabricmanager` — get upgraded out of sync with the driver.
- In-place image refresh (`REFRESH_MODE=true`) on Ubuntu was inheriting the Lustre kernel hold + GRUB stickiness logic, which defeats the whole point of a refresh.

### Changes

**EL9.x version bump (Alma/Rocky 9.7 → 9.8)**
- versions.json: rename all `almalinux9.7` / `rocky9.7` keys to `9.8`, bump DOCA `3.2.1` → `3.2.2` for the rhel9 host package.
- variables.pkr.hcl: bump `distro_version` defaults, kernel map (`9.8 = 5.14`), marketplace SKU map, and SIG image definition name to `AlmaLinuxHPC-9.8` / `Rocky-9.8`.
- Folder consolidation: rename distros/almalinux9.7/ → almalinux9.x and distros/rocky9.7/ → rocky9.x. EL9 install scripts are not minor-specific, so a new local in variables.pkr.hcl resolves any `alma|rocky 9.*` distro version to the shared `9.x` folder.

**In-place refresh kernel-hold fix**
- build.pkr.hcl: propagate `REFRESH_MODE=${local.refresh_mode}` to both the prerequisites and install provisioners.
- prerequisites.sh: gate every `apt-mark hold` / `grub-set-default` / `GRUB_SAVEDEFAULT` on `LUSTRE_BUILD_FROM_SOURCE != true && REFRESH_MODE != true` (7 call sites, GB200 + LTS 22.04 + LTS 24.04 paths).

**New `dnf_pin_packages` helper**
- utilities.sh: adds an idempotent `dnf_pin_packages` that maintains a single `exclude=` line in `/etc/dnf/dnf.conf`, creating it if absent, deduplicating, and rejecting empty arguments.
- Replaces the fragile `sed -i "$ s/$/ PKG/"` pattern in:
  - install_gdrcopy.sh — `gdrcopy*`
  - install_lustre_client.sh — `amlfs*`
  - install_mpis.sh — `ucx*`, `openmpi`, `perftest`
  - install_nccl.sh — `libnccl*` (both RHEL and AzureLinux paths)
  - install_nvidia_container_toolkit.sh — `*nvidia-container*` (also drops the redundant `yum update -y` immediately after the repo add)
  - install_nvidia_fabric_manager.sh — `${PACKAGE_NAME}`

**DOCA / CUDA repo-local excludes (EL9.8 compatibility)**
- install_doca.sh: after `dnf -y install doca-ofed`, add a `baseos` repo-local exclude (intentionally not a global `/etc/dnf/dnf.conf` pin, so `dnf check-update` in `verify_package_updates` still flags real updates):
  - `baseos.excludepkgs=rdma-core* libibverbs* libibumad* librdmacm* libibmad* libhns* perftest infiniband-diags python3-pyverbs iwpmd srp_daemon`
  - AlmaLinux/Rocky 9.8 baseos shipped a brand-new RDMA-core stack with a separate `libhns` provider and ABI-bumped `libibverbs`/`perftest` (IBVERBS_1.15 / HNS_1.0 symbols) incompatible with DOCA's `libibverbs-2510.0.11-1.el9`. Without this exclude, `install_pmix.sh`'s `yum update -y` aborts with `cannot install both libibverbs-61.0-2.el9 from baseos and libibverbs-2510.0.11-1.el9 from @System`. DOCA's userland repo is left as the only rdma-core provider at runtime.
- install_nvidiagpudriver.sh: exclude `mft*` and `kernel-mft*` from the `cuda-rhel9` repo so `mft` continues to track DOCA (which built the matching `kernel-mft-dkms`) instead of CUDA — applied at the point `cuda-rhel9.repo` is added (which is why it lives here and not in `install_doca.sh`).

**AlmaLinux / Rocky 8.10 moby install fix**
- install_utils.sh and install_utils.sh: inject `module_hotfixes=1` into the downloaded `microsoft-prod.repo`. Microsoft's `moby-runc` rpm `Provides: runc`, which is otherwise hidden by the `container-tools` AppStream module and broke `moby-engine` installation.

**NVSwitch / Fabric Manager test-time fixes**
- test-definitions.sh: new `sku_has_nvswitch` helper covering NDv4 A100 and NDv5 H100/H200; `verify_nvidia_fabricmanager_service` and `verify_sku_customization_service` now use it. Fixes a latent bug where the regex `standard_nd96is*_h100_v5` was a shell glob, not ERE, and never matched H200 v5.
- run-tests.sh: new `ensure_nvidia_fabricmanager_active` invoked before the NVIDIA test matrix. On NVSwitch SKUs `cuInit()` returns `CUDA_ERROR_SYSTEM_NOT_READY` until FM finishes NVLink fabric setup, which was failing `gdrcopy_sanity` during build-time validation. Uses `sudo -n systemctl` because the script runs as the non-root Packer build user.


arsdragonfly (dismissed):

utils/utilities.sh
```diff
@@ -134,3 +134,60 @@ function sku_has_infiniband {
...
+        new="${new# }"
+        if [[ "$new" != "$current" ]]; then
+            # '|' is safe as a sed delimiter: package globs contain only
+            # [A-Za-z0-9._*-], never '|' or '\'.
+            sed -i "s|^exclude=.*|exclude=${new}|" /etc/dnf/dnf.conf
+        fi
+    else
+        echo "exclude=${pkgs[*]}" >> /etc/dnf/dnf.conf
+    fi
+}
```
ColtonPaul: Dropped the final newline

utils/utilities.sh
```diff
@@ -134,3 +134,60 @@ function sku_has_infiniband {
...
+#             Empty-string arguments are rejected: callers must not pass
+#             unset variables or empty array expansions. There is no
+#             legitimate "pin nothing" use case -- such a call almost
+#             always indicates a bug (e.g. unset $PACKAGE_NAME or an
+#             empty mapfile result) and would silently corrupt the
+#             'exclude=' line with stray whitespace.
+# @Returns  : 0 on success, 0 (no-op) if /etc/dnf/dnf.conf is absent,
+#             1 if no non-empty package arguments were provided.
+############################################################################
+function dnf_pin_packages {
```
ColtonPaul, line 162: What would you think about adding a new test on builds to validate that dnf.conf looks as expected (i.e. that it's not corrupted and has a list of packages on the exclude line that you'd expect to see)? It sounds like it failed silently before your fix, so maybe we can make it louder if something breaks it again.

components/install_doca.sh
```diff
@@ -84,28 +84,56 @@ EOF
...
     dnf -y install doca-ofed
 else
     # RHEL-family: AlmaLinux, Rocky Linux, RHEL, etc.
     rpm -i $DOCA_FILE
     dnf clean all
-    
-    # Install DOCA extras for compatibility
-    dnf install -y doca-extra
-    
-    /opt/mellanox/doca/tools/doca-kernel-support
```
ColtonPaul, line 99: Just for my education, why aren't these needed anymore?

normanvuong: Oh this is not needed anymore in 3.x DOCA: 
image

utils/utilities.sh
normanvuong: I have added the appropriate test. Lmk if you have further questions!

ColtonPaul (approved):

PR #555, opened 2026-06-11
Bump Lustre to 2.17 on U24 and switch to DKMS on RHEL

arsdragonfly:


distros/almalinux8.10/install_utils.sh
```diff
@@ -92,14 +92,14 @@ rm -f environment-modules-5.0.1-1.el9.x86_64.rpm
...
-## kernel upgrades, matching the Ubuntu prerequisites.sh behavior). The
+## Disable kernel updates (skipped when refreshing an image in-place so that
+## DKMS-style rebuilds can keep up with kernel upgrades, matching the
+## Ubuntu prerequisites.sh behavior). The
 ## shim*/grub2* sed lines extend the just-added exclude= directive, so they
 ## must stay inside the same conditional -- running them without an exclude
 ## line would corrupt the previous last line of dnf.conf (e.g.
 ## skip_if_unavailable=False).
-if [[ "${LUSTRE_BUILD_FROM_SOURCE,,}" != "true" && "${REFRESH_MODE,,}" != "true" ]]; then
+if [[ "${REFRESH_MODE,,}" != "true" ]]; then
```
normanvuong: Should we remove pinning the kernel since now dkms support all of our distros?

arsdragonfly: good catch, will remove the pinning later

arsdragonfly: pinning removed

normanvuong (approved):

distros/azurelinux3.0/install_utils.sh
```diff
@@ -88,17 +88,6 @@ tdnf install -y numactl \
     dos2unix \
     azcopy
 
-# Disable kernel updates
-echo "exclude=kernel* kmod*" | tee -a /etc/dnf/dnf.conf
```
normanvuong, line 92: Just for my own curiosity. Why did azlinux3 has a kernel hold when it doesn't have to install lustre?

arsdragonfly: It was cargo-culted from other distros

PR #562, opened 2026-06-19
eliminate tmp working dir and pin dynolog commit

arsdragonfly:
per AI investigation of code, working dir doesn't need to be writable (systemd defaults to "/")

### DIFF
--- a/components/install_dynolog_drl.sh
+++ b/components/install_dynolog_drl.sh
@@ -45,9 +45,10 @@ if [[ "$GPU" == "NVIDIA" ]]; then
     ##########################################################################
     # Build and install dynolog
     ##########################################################################
-    git clone --recurse-submodules -j8 --branch v${DYNOLOG_VERSION}  $DYNOLOG_URL /tmp/dynolog
+    git clone --no-checkout ${DYNOLOG_URL} /tmp/dynolog
     pushd /tmp/dynolog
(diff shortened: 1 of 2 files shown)


components/install_dynolog_drl.sh
```diff
@@ -14,10 +14,12 @@ if [[ "$GPU" == "NVIDIA" ]]; then
 
     dynolog_metadata=$(get_component_config "dynolog")
     DYNOLOG_VERSION=$(jq -r '.version' <<< $dynolog_metadata)
+    DYNOLOG_COMMIT=$(jq -r '.commit' <<< $dynolog_metadata)
```
ljqqqqq, line 17: Curious why should we pin the commit?

darylmsft: from claude.4.8:
pinning to an immutable SHA prevents a moved/re-tagged v1.0.0 from silently changing what gets built (matches the repo's gdrcopy/rccl convention), which is the integrity half of this fix

ljqqqqq: Understood, thanks Daryl!

darylmsft (approved):
RTM (yutongs) accepts the technical debt of taking a hard dependency on a personal github account.

normanvuong (approved):

PR #566, opened 2026-06-24
Jiaqi/packer bm

ljqqqqq:
### Build Targeting Flag

- Replaces the older `TARGET_IMAGE_VARIANT` / `NODE_TYPE` model with `TARGET_NODE_TYPE`.
   - Adds supported target types: `azure_vm_regular`, `azure_vm_akshost`, `baremetal_1p`, and `baremetal_3p`.
   - Reorganizes `versions.json` so component lookup is keyed by distro -> arch -> target node type -> sku
   - Normalize the usage of DISTRIBUTION. Moves AKS-specific Ubuntu 24.04 metadata under `azure_vm_akshost` instead of using `ubuntu24.04-aks`. 
   -  Reorganizes `tests/test-matrix_NVIDIA.json` so tests matrix are split under:
      - `azure_vm_regular`
      - `baremetal_3p`
      - `azure_vm_akshost`
      - baremetal_1p tests are added by overlay files

### Baremetal 1P support in Packer:
  - uses specialized image mode for `baremetal_1p`
  - switches the default SSH user to `azhpcuser`
  - passes ADO/login secrets into the install flow
  - **uploads baremetal overlay files** for MRC-specific scipts.
  -  **Adds override hook loading through `/opt/azurehpc/test/test-overrides.sh`.**
  - Adds MRC / baremetal 1P handling for non-MRC-specific components which are shared by MRC nodes but the steps are slightly different.

### Install Flow

- Deletes the GB200-only Ubuntu 24.04 NVIDIA driver script and routes GB200, AKS, and regular Ubuntu 24.04 through the common `components/install_nvidiagpudriver.sh` path.
- Treats VR200 as a temporary GB200 install path in Ubuntu 24.04 distro install logic.
- Changes DOCA, NCCL, MPI, and test logic from simple InfiniBand checks to a network mode model: `no_rdma`, `mrc`, or `standard_ib` to decouple the relationship between SKU and network stack selection. 
- Generalizes internal bits download from GB200-specific variables to `INTERNAL_BITS_CONTAINER_NAME` and `INTERNAL_BITS_BLOB_NAME`.


components/install_gdrcopy.sh
```diff
@@ -52,7 +52,7 @@ else
         git checkout ${GDRCOPY_COMMIT}
         if [[ $DISTRIBUTION == *"ubuntu"* ]]; then
             # Install gdrcopy
-            apt install -y build-essential devscripts debhelper check libsubunit-dev fakeroot pkg-config dkms
+            apt install -y build-essential devscripts debhelper check libsubunit-dev fakeroot pkg-config
```
normanvuong, line 55: why are we removing dkms installation?

distros/almalinux8.10/install.sh
```diff
@@ -78,6 +78,9 @@ rm -rf /var/intel/
 # optimizations
 $COMPONENT_DIR/hpc-tuning.sh
 
+# install Azure Linux Agent
+$COMPONENT_DIR/install_waagent.sh
```
normanvuong, line 82: I think hpc-tuning.sh already run install_waagent.sh so this might be duplication?

ljqqqqq: The calling of install_waagent.sh in hpc-tuning.sh has been removed to accommodate the needs of baremetal_1p, which does not need waagent but needs to share hpc-tuning.sh

components/install_gdrcopy.sh
ljqqqqq: Because mrc images use privately shared nvidia repo and set its priority to the highest. https://github.com/ljqqqqq/azhpc-images/blob/jiaqi/packer-bm/components/install_nvidiagpudriver.sh#L66

Dkms has already been installed in install_utils.sh. The local NVIDIA repository also contains a dkms package. If we install dkms here again, APT will prefer the version from the local NVIDIA repository because of its priority (1001), causing the existing dkms package to be replaced and generating a false warning of package downgrade

Shuyu-Zhao (dismissed):

MKGAURAB (approved):

normanvuong (approved):

PR #567, opened 2026-06-26
patch DOCA on RHEL family and rip out o2ib from Lustre

arsdragonfly:
DOCA-OFED kmod fails to install on RHEL family due to kernel API change and needs DKMS patching
Lustre on Azure has no IB line-of-sight, so disabling o2ib to avoid compilation errors

### DIFF
--- a/components/install_doca.sh
+++ b/components/install_doca.sh
@@ -99,6 +99,7 @@ EOF
     rm -f /tmp/hpcx-provides-openmpi_*_all.deb /tmp/hpcx-provides-openmpi
 
     apt-get -y install doca-ofed
+    check_dkms_status mlnx-ofed-kernel iser isert srp
 elif [[ $DISTRIBUTION == "azurelinux3.0" ]]; then
     rpm -i $DOCA_FILE
     dnf clean all
@@ -113,6 +114,7 @@ else
     sed -i '/^exclude=/d' /etc/dnf/dnf.conf
     configure_mlnx_ofa_kernel_dkms_dpll_patch
     dnf -y install doca-ofed
+    check_dkms_status mlnx-ofa_kernel iser isert srp
     # Restore exclusion
     mv /etc/dnf/dnf.conf.bak /etc/dnf/dnf.conf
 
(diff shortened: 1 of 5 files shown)


components/install_doca.sh
```diff
@@ -11,6 +11,24 @@ DOCA_FILE=$(basename ${DOCA_URL})
 
 download_and_verify $DOCA_URL $DOCA_SHA256
 
+configure_mlnx_ofa_kernel_dkms_dpll_patch() {
+    local kernel_header=/usr/src/kernels/$(uname -r)/include/linux/dpll.h
+    local dkms_conf=/etc/dkms/mlnx-ofa_kernel.conf
+    local patch_file=${azhpc_dir}/components/patches/mlnx-ofa-kernel-dpll-ffo-param.patch
```
normanvuong: Shouldn't this be: local patch_file=${COMPONENT_DIR}/patches/mlnx-ofa-kernel-dpll-ffo-param.patch

normanvuong (changes requested):

components/install_doca.sh
arsdragonfly: fixed

normanvuong (dismissed):

components/install_lustre_client.sh
```diff
@@ -7,6 +7,64 @@ source ${UTILS_DIR}/utilities.sh
...
+configure_legacy_lustre_dkms_no_o2ib() {
+    local module=lustre-client-modules
+    local module_version=$1
+    local dkms_conf=/etc/dkms/${module}-${module_version}.conf
+
+    mkdir -p /etc/dkms
+    cat > "${dkms_conf}" <<'EOF'
+# Azure clients do not have IB line of sight to Lustre servers, so use TCP LNet.
+# Ubuntu 22.04's published AMLFS Lustre DKMS package has a static dkms.conf with
+# ko2iblnd in the expected module list, so override both configure and artifacts.
```
normanvuong: is there a good way to check after installation that lustre was build with no o2ib? I'm just thinking there might be cases where the patch silently failed/performed no-op and the image actually still get build with o2ib.

arsdragonfly: it's more about saving build time and avoiding unnecessary build errors

normanvuong (approved):

PR #568, opened 2026-06-26
Restore and harden lustre check

ColtonPaul:
The Lustre functional check was disabled in the past to not block experimental and/or builds on brand new kernels that cannot install Lustre, but this check is necessary as a correctness gate and needs to be restored. In addition to checking the lustre client, verify the kernel module is listed as well

### DIFF
--- a/tests/test-definitions.sh
+++ b/tests/test-definitions.sh
@@ -455,16 +455,20 @@ function verify_ib_modules_and_devices {
     fi
 }
 
-# only best-effort install since Lustre isn't always available
-# function verify_lustre_installation {
-#     case ${ID} in
-#         ubuntu) dpkg -l | grep lustre-client;;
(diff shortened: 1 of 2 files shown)


tests/test-definitions.sh
```diff
@@ -455,16 +455,20 @@ function verify_ib_modules_and_devices {
     fi
 }
```
arsdragonfly, line 457: exempt AZL3

arsdragonfly (changes requested):

tests/test-definitions.sh
ColtonPaul: AZL3 is implicitly skipped by not having check_lustre in its test-matrix json

arsdragonfly (approved):

PR #591, opened 2026-08-12
Simplify in-place refresh component script

normanvuong:
## Summary

- Limit in-place manifest refresh to components that can change during `apt upgrade` or `dnf update`.
- Preserve install-time versions for pinned and source-installed components.
- Record the active kernel after reboot.
- Keep APT’s native DKMS hook responsible for rebuilding modules in dependency order.
- Simplify refresh logic by removing broad best-effort detection and strict tier handling.

## Validation
https://dev.azure.com/hpc-platform-team/hpc-image-val/_build/results?buildId=36810&view=results

### DIFF
--- a/components/refresh_component_versions.sh
+++ b/components/refresh_component_versions.sh
@@ -4,13 +4,13 @@ set -euo pipefail
 # =============================================================================
 # Refresh Component Versions
 # =============================================================================
-# Regenerates /opt/azurehpc/component_versions.txt by detecting installed
-# versions of all HPC components on the running system.
+# Refreshes /opt/azurehpc/component_versions.txt for components whose versions
+# can change during the prerequisite apt/dnf transaction.
 #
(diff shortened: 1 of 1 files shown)


ColtonPaul:
The simplification is good, but it does seem like we're losing the confidence that we have a "detector" for components that can upgrade during an in-place upgrade. Is there a compromise option where we get the best of both worlds:
1. We don't have to write version discovery code for components that can't upgrade
2. If the list of upgradeable components in any component_versions.json changes, we're not at risk of forgetting to update refresh_component_versions.sh accordingly?

I'm curious about something like an `in-place-upgradeable` attribute in versions.json that could help solve this and possibly also help lock down intentions about which package versions can move (for example, I see that pmix still has a detector, so I checked the code and saw that only Ubuntu holds it, and RHEL doesn't. Is that intentional?).

Issue #585, opened 2026-07-24
Add `datacenter-gpu-manager-4-multinode-*` to images


yosoyjay:
A standard mutli-node and rack-level test `mnubergemm` is now included in the packages `datacenter-gpu-manager-4-multinode-*` eliminating the need for users to manually install via other methods.  The ask is to include this package in the AzHPC images by default to avoid having users and operators from manually installing.

normanvuong:
The PR has been merged to add the datacenter-gpu-manger-4-mulitnode-* and it should be included in September image refresh: https://github.com/Azure/azhpc-images/pull/600

PR #600, opened 2026-08-26
Add datacenter-gpu-manager-4-multinode-* to images

normanvuong:
A standard mutli-node and rack-level test mnubergemm is now included in the packages datacenter-gpu-manager-4-multinode-* eliminating the need for users to manually install via other methods. The ask is to include this package in the AzHPC images by default to avoid having users and operators from manually installing.
https://github.com/Azure/azhpc-images/issues/585


ljqqqqq:
Suppose this should wait for the test result on validation pipeline

Shuyu-Zhao (approved):

normanvuong:
> Suppose this should wait for the test result on validation pipeline

yes I've tested and all the benchmark passed successfully on all the current SKU + OS combination that we had.

PR #601, opened 2026-08-26
introduce extra args for rebuild parity with original HPC-X to improv…

arsdragonfly:
…e robustness


v-hunter-ms (approved):
Reviewed the code changes. The common OpenMPI configure options are applied across the rebuild paths, while the Mellanox optimized platform option remains limited to the UCX paths. The --cuda argument is conditionally added only for NVIDIA builds, and all branches continue to use the external PMIx configuration. I did not identify any blocking issues in the code changes.

normanvuong (dismissed):

normanvuong (approved):

PR #608, opened 2026-09-13
Harden MDATP cleanup and Ubuntu build reliability

ljqqqqq:
## Summary

This PR addresses two image-build issues:

- Hardens Microsoft Defender for Endpoint (MDE/MDATP) cleanup so platform re-provisioning cannot leave Defender residue in the captured image.
- Updates the Ubuntu Dynolog build toolchain to Rust/Cargo 1.89 so current Cargo dependencies can compile.

## Background

MDE can be installed automatically through the `MDE.Linux` extension while an image is being built. A single early `mdatp` purge is not sufficient: the Azure Guest Agent can reconcile the extension again during the gap between cleanup and image capture. That can leave the package, onboarding/configuration data, install logs, systemd state, extension artifacts, or EICAR test files in the final image and cause LISA/Marketplace validation failures.

The Dynolog build was also using Rust 1.82, while its resolved `textwrap 0.16.4` dependency requires Rust 1.88 or newer.

### MDATP cleanup

- Stops `walinuxagent`/`waagent` before the initial cleanup and again in the epilog to prevent MDE extension reconciliation during the capture window.
- Purges both `mdatp` and `microsoft-mdatp` package states on Ubuntu and RPM-based images.
- Repeats the purge immediately before image capture as a final backstop.
- Removes MDE-specific program, state, configuration, onboarding, log, audit, AppArmor, systemd, dpkg metadata, and extension residue.
- Removes EICAR test artifacts (`eicar.com*` and `utils_58.py`) without crossing into mounted data disks.
- Reloads systemd after removing unit files.
- Adds `verify_final_cleanup`, which blocks image capture if package state, dpkg metadata, the executable or service, a running `wdavdaemon`, MDE filesystem/extension residue, or EICAR artifacts remain.

Cleanup is intentionally limited to MDE-specific paths rather than broad Microsoft parent directories that may contain files owned by other products.

### Dynolog

- Installs Rust and Cargo 1.89 on Ubuntu.
- Uses a single `RUST_VERSION` value for package selection and the toolchain `PATH`.

## Scope

This PR no longer changes direct `dpkg -i` calls or dpkg/APT lock handling.

## Validation

- `git diff --check` passes.
- Bash syntax validation passes for the three changed scripts.
- The final PR diff is limited to the two MDATP cleanup scripts and the Dynolog installer.

### DIFF
--- a/utils/clear_history_epilog.sh
+++ b/utils/clear_history_epilog.sh
@@ -62,11 +62,155 @@ find_azurelinux_distro() {
 distro=`find_distro`
 echo "Detected distro: ${distro}"
 
+verify_final_cleanup() {
+    local path
+
+    if [[ ${distro} == *"Ubuntu"* ]] && \
+        dpkg -l 2>/dev/null | grep -qE '^(ii|rc|hi|ri|pi|ip|in)[[:space:]]+(mdatp|microsoft-mdatp)(:|[[:space:]])'; then
+        echo "ERROR: mdatp has a dpkg state before image capture"
+        dpkg -l 2>/dev/null | grep -E '(mdatp|microsoft-mdatp)' || true
+        return 1
+    elif [[ ${distro} != *"Ubuntu"* ]] && {
+        rpm -q mdatp >/dev/null 2>&1 || rpm -q microsoft-mdatp >/dev/null 2>&1
+    }; then
+        echo "ERROR: mdatp package is still present before image capture"
+        return 1
+    fi
+
+    if [[ ${distro} == *"Ubuntu"* ]] && {
+        compgen -G '/var/lib/dpkg/info/mdatp.*' >/dev/null ||
+            compgen -G '/var/lib/dpkg/info/microsoft-mdatp.*' >/dev/null
+    }; then
+        echo "ERROR: mdatp dpkg metadata is still present before image capture"
+        ls -la /var/lib/dpkg/info/mdatp.* /var/lib/dpkg/info/microsoft-mdatp.* 2>/dev/null || true
+        return 1
+    fi
+
+    if command -v mdatp >/dev/null 2>&1; then
+        echo "ERROR: mdatp executable is still present before image capture"
+        return 1
+    fi
+
+    if command -v systemctl >/dev/null 2>&1 && {
+        systemctl cat mdatp.service >/dev/null 2>&1 ||
+            systemctl list-unit-files 2>/dev/null | grep -qw mdatp.service
+    }; then
+        echo "ERROR: mdatp.service is still present before image capture"
+        systemctl status mdatp.service 2>/dev/null || true
+        return 1
+    fi
+
+    if pgrep -x wdavdaemon >/dev/null 2>&1; then
+        echo "ERROR: wdavdaemon is still running before image capture"
+        pgrep -a -x wdavdaemon || true
+        return 1
+    fi
+
+    for path in \
+        /opt/microsoft/mdatp \
+        /opt/microsoft/mde \
+        /var/opt/microsoft/mdatp \
+        /etc/opt/microsoft/mdatp \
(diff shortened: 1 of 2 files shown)


normanvuong (dismissed):

utils/utilities.sh
```diff
@@ -116,6 +116,53 @@ verify_checksum() {
...
+                return "${status}"
+            fi
+        fi
+
+        echo "Waiting for dpkg lock; retrying dpkg -i in ${retry_interval_seconds}s" >&2
+        : > "${stderr_file}"
+        sleep "${retry_interval_seconds}"
+    done
+}
+
```
arsdragonfly, line 165: we already have lock timeout config for apt [here](https://github.com/Azure/azhpc-images/blob/89f89358450d509d7d9bb499df7ba92c4a44473d/packer/scripts/prerequisites.sh#L33), switch `dpkg -i` to `apt install -y` instead of using this wrapper

utils/clear_history_epilog.sh
```diff
@@ -62,12 +62,156 @@ find_azurelinux_distro() {
...
+elif [[ ${distro} == *"Azure Linux"* ]]; then
+    if rpm -q mdatp >/dev/null 2>&1; then
+        dnf remove -y mdatp
+    fi
+else
+    if rpm -q mdatp >/dev/null 2>&1; then
+        dnf remove -y mdatp
+    fi
+fi
+
```
arsdragonfly, line 174: coalesce these branches

arsdragonfly (changes requested):

utils/utilities.sh
```diff
@@ -116,6 +116,53 @@ verify_checksum() {
...
+    local now
+    local elapsed
+    local status
+    local stderr_file
+
+    start_time=$(date +%s)
+    stderr_file=$(mktemp)
+
+    while true; do
+        if dpkg -i "$@" 2> >(tee "${stderr_file}" >&2); then
```
YutongSun: If tee hasn't completed the file before dpkg returns, grep may miss the lock error in the file. Suggest sync redirect. E.g.

if dpkg -i "$@" 2> "${stderr_file}"; then
    cat "${stderr_file}" >&2
    rm -f "${stderr_file}"
    return 0

utils/utilities.sh
ljqqqqq: The DPkg::Lock::Timeout option is in APT configuration and can only be used on APT. We still meet dpkg lock issue when running with dpkg -i. For example: https://dev.azure.com/hpc-platform-team/hpc-image-val/_build/results?buildId=38059&view=logs&j=2e985d55-ce88-5f0d-2e8a-908bf0ea94b7&t=ea2e6eb5-7ade-5086-6c60-7000dc5145bb&l=18376. And I also meet several similar issues on gdrcopy previously

utils/clear_history_epilog.sh
ljqqqqq: https://github.com/ljqqqqq/azhpc-images/blob/456c95e78141cd8b6a109e081f192a61ee37914c/utils/clear_history.sh

I kept the style same with other codes. I suppose the branch dividing makes sense?

arsdragonfly: it was me being dumb and forgetting to remove redundant branches after [dnf alignment](https://github.com/Azure/azhpc-images/pull/594), but in principle we shouldn't have separate branches when it's not necessary

utils/utilities.sh
arsdragonfly: `apt install -y` should be able to install .deb packages directly while handling locks gracefully

ljqqqqq: I think we use dpkg -i intentionally, which can install the exact package like doca meta package and preserve the installation orders. Changing all "dpkg -i" to "apt install -y" needs much more test and are out of scope of this PR

arsdragonfly (approved):

PR #611, opened 2026-09-14
Install explicit ROCm package list without MIVisionX

normanvuong:
MIVisionX introduce various CVE regarding FFmepg, Qt, cjson and mbedTLS so install explicit ROCm package list without MIVisionX to avoid those CVE instead of using Ubuntu Pro


components/install_rocm.sh
```diff
@@ -20,7 +20,15 @@ if [[ $DISTRIBUTION == *"ubuntu"* ]]; then
...
-        amdgpu-install -y --usecase=graphics,rocm
+        amdgpu-install -y --usecase=graphics
+        apt-get install -y \
+            rocm-utils \
+            rocm-developer-tools \
+            rocm-openmp-sdk \
+            rocm-opencl-sdk \
+            rocm-ml-sdk \
+            migraphx migraphx-dev \
+            rpp rpp-dev
```
arsdragonfly, line 36: add TODO comment to give more context

arsdragonfly (changes requested):

components/install_rocm.sh
normanvuong: Added! Thanks!

PR #621, opened 2026-09-18
fix ROCm-enabled HPC-X rebuild on Ubuntu 26.04

normanvuong:
Bug 1:
add ROCm ld.so.conf entry on Ubuntu 26.04

The amdrocm-core-sdk packaging installs into /opt/rocm/core-10.0/lib and
exposes /opt/rocm/lib through update-alternatives, but unlike the
amdgpu-install/rocm metapackage used on 22.04 and 24.04 it ships no
/etc/ld.so.conf.d entry. libamdhip64.so.7 was therefore unresolvable,
breaking the Open MPI rebuild at link time ("libamdhip64.so.7, needed by
libopen-pal.so, not found") and leaving libopen-pal unloadable at runtime.

Bug2:
build Open MPI against its internal PMIx, hwloc and libevent

HPC-X's ompi5/lib holds the bundled PMIx, hwloc and libevent alongside a
complete vendor Open MPI. Passing that directory to --with-pmix,
--with-hwloc and --with-libevent put it ahead of hpcx-rebuild/lib in every
RUNPATH, so the rebuilt binaries loaded the stock libmpi and libopen-pal
instead of ours -- the ROCm accelerator, the ROCm-enabled UCX and the HCOLL
rebuild were all silently inert. The deprecated standalone Open MPI build
had the same defect.

Open MPI ships PMIx, PRRTE, hwloc and libevent under 3rd-party/ and the
HPC-X source tarball carries them fully populated, so =internal builds them
into Open MPI's own prefix and no vendor path reaches the link line at all.
RUNPATH is now a single entry and libmpi, libopen-pal, libpmix, libhwloc
and libevent all resolve under it.

Assert after the rebuild that ompi_info resolves libmpi under
hpcx-rebuild/lib so this cannot regress silently, and report the PMIx
version from the tree actually shipped.

Bug 3:
fix HPC-X rebuild modulefile pointing at a missing directory

hpcx_rebuild.sh generates the module by rewriting the vendor template with
sed "s;hpcx_dir/ompi[0-9]*;hpcx_dir/$name;". The template spells the path
$hpcx_dir/ompi$mpi_version, so ompi[0-9]* matches only "ompi" and the
substitution leaves the suffix behind as $hpcx_dir/hpcx-rebuild$mpi_version.
That directory is never created, so module load mpi/hpcx exported a bogus
MPI_HOME, PATH, LD_LIBRARY_PATH and PKG_CONFIG_PATH.

Drop the dangling suffix from the generated module and init script. Only
that one usage is rewritten; mpi_version is still needed for the
ompi$mpi_version/tests path. The ompi_prefix env variable does not help
here, since $name is substituted in ahead of the suffix either way.

Affects every distro that rebuilds Open MPI 5, where the wrong module would
also have shipped in the image and broken mpi/hpcx for end users.

Bug 4:
pass a defined HPC-X prefix to the rccl-tests build

CMAKE_PREFIX_PATH referenced $HPCX, which is never set, so the configure
line ended in a bare separator and HPC-X was absent from the search path.
The hpcx module exports HPCX_MPI_DIR, so use that.


components/install_mpis.sh
```diff
@@ -57,11 +57,34 @@ else
...
-    PMIX_PATH=${HPCX_PATH}/ompi5
-    HPCX_OMPI5_PKG_CONFIG_PATH=${PMIX_PATH}/lib/pkgconfig
+    HPCX_OMPI5_PATH=${HPCX_PATH}/ompi5
+    HPCX_OMPI5_PKG_CONFIG_PATH=${HPCX_OMPI5_PATH}/lib/pkgconfig
     PKG_CONFIG_PATH=${HPCX_OMPI5_PKG_CONFIG_PATH} pkg-config --exists 'pmix >= 5' hwloc libevent
     PMIX_VERSION=$(PKG_CONFIG_PATH=${HPCX_OMPI5_PKG_CONFIG_PATH} pkg-config --modversion pmix)
     write_component_version "PMIX" "${PMIX_VERSION}"
+
+    # ompi5/lib also ships the vendor libmpi/libopen-pal. Passing that directory to
+    # --with-pmix/--with-hwloc/--with-libevent puts it ahead of hpcx-rebuild/lib in every
```
arsdragonfly: Could you try with-pmix/hwloc/libevent=internal and see if it builds them from source?

arsdragonfly (changes requested):

components/install_mpis.sh
normanvuong: ack, testing this out rn.

normanvuong: work on a live box, I updated to use that option instead.

arsdragonfly (approved):

